PR #79, opened 2022-11-06
Adding cloud reconciler and supprting integration with Otterize cloud

evyatarmeged:
### Description
Adding a cloud reconciler to integrate with Otterize cloud. Should run only when client id & client secret were provided (true for operator & watcher).
Added cloud addr as environment variable (will be added to helm with a default value).
The repository now includes GQL schemes to enable communication with the cloud.


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "helm-charts"]
 	path = helm-charts
 	url = https://github.com/otterize/helm-charts.git
+[submodule "api"]
```
orishoshan, line 4: Change the GHA `checkout` action so that it also checks out submodules during build

src/operator/api/v1alpha1/intents_types.go
```diff
@@ -152,6 +155,30 @@ func init() {
 	SchemeBuilder.Register(&ClientIntents{}, &ClientIntentsList{})
 }
 
+func (in *ClientIntentsList) FormatAsOtterizeIntents() ([]*intents.IntentInput, error) {
+	otterizeIntents := make([]*intents.IntentInput, 0)
+	for _, clientIntents := range in.Items {
+		for _, intent := range clientIntents.GetCallsList() {
+			b, err := json.Marshal(intent)
```
orishoshan, line 162: This is a neat trick but it has the side effect of breaking silently if the fields change - some of them will just be ignored. This is losing a lot of the power of having these two separate APIs (GraphQL and Kubernetes) have generated statically typed code.

If you map the fields with code instead, the code will fail to compile if the structs change. Please do so - it will help those that come after you make changes easily and have confidence that they will work.

src/operator/controllers/intents_reconcilers/otterize_cloud/cloud_reconciler.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (r *OtterizeCloudReconciler) ApplyIntentsToCloud(
+	ctx context.Context,
+	intentsInput []*intents.IntentInput,
+	ns string) (ctrl.Result, error) {
+
+	fmt.Println("yalla intents")
+	c := r.newClientForURI(ctx, "intents/query")
+
+	// TODO: Change this intents source to be meaningful. Consider using integration name
```
orishoshan, line 90: Should complete this TODO before finishing this PR - using the integration info sounds good, like you wrote

src/operator/controllers/intents_reconcilers/otterize_cloud/cloud_reconciler.go
```diff
@@ -0,0 +1,98 @@
...
+		fmt.Sprintf("%s/%s", r.cloudAddr, uri),
+		oauth2.NewClient(ctx, r.tokenSrc))
+}
+
+func (r *OtterizeCloudReconciler) ApplyIntentsToCloud(
+	ctx context.Context,
+	intentsInput []*intents.IntentInput,
+	ns string) (ctrl.Result, error) {
+
+	fmt.Println("yalla intents")
```
orishoshan, line 87: yalla! oopsie :)

src/operator/controllers/intents_reconcilers/otterize_cloud/cloud_reconciler.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func NewOtterizeCloudReconciler(
+	client client.Client,
+	scheme *runtime.Scheme,
+	otterizeClientID,
+	otterizeClientSecret,
+	cloudAddr string) *OtterizeCloudReconciler {
+
+	cfg := clientcredentials.Config{
```
orishoshan, line 35: This bit of code is copypasted here and in NamespaceWatcher - factor it out to a function that returns `graphql.Client` and use that instead.

A nice trick you can do is add `viper` configuration, bind it to pflags, and then have that function use the client ID, client secret and cloud addr from the configuration directly, rather than passing it around (since they are global - the same for the operator no matter where the code is placed). This way the function will only need to take the URI subpath, such as `intents/query`.

Check out `GetAPITokenSource` in the old CLI, and the `viper` docs: https://pkg.go.dev/github.com/spf13/viper#readme-working-with-flags

src/watcher/reconcilers/namespace.go
```diff
@@ -47,6 +72,10 @@ func (ns *NamespaceWatcher) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	if ns.tokenSrc != nil {
```
orishoshan, line 75: Remove this if it's not implemented in this PR

orishoshan (changes requested):

src/operator/controllers/intents_reconcilers/otterize_cloud/cloud_reconciler.go
evyatarmeged: Will not happen until some discussions are resolved. We can leave this PR out for now.

src/operator/api/v1alpha1/intents_types.go
evyatarmeged: Alright agree, will revert (it was like this originally). 
Just FYI I've aligned cloud intents to use the same format as the k8s intents. Should I keep the changes or revert them (I prefer the formats to be aligned but it doesn't matter if the conversion is code-based).

src/watcher/cmd/main.go
```diff
@@ -27,20 +28,25 @@ func main() {
...
 		"Namespaces that will be watched by the operator. Specify multiple values by specifying multiple times or separate with commas.")
+	pflag.StringVar(&otterizeClientID, "client-id", "", "Otterize cloud Kubernetes integration client id")
+	pflag.StringVar(&otterizeClientSecret, "client-secret", "", "Otterize cloud Kubernetes integration client secret")
 
 	pflag.Parse()
 
 	ctrl.SetLogger(logrusr.New(logrus.StandardLogger()))
 
 	var err error
+	cloudAddr := os.Getenv("OTTERIZE_API_ADDRESS") // we use Getenv because empty string is viable
```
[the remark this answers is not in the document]
evyatarmeged: I thought about having this as an env var with a default value would make the most sense since the address is unlikely to be changed that much. (disregard comment, it's copied from someplace else and I'll delete it).

src/watcher/cmd/main.go
```diff
@@ -72,7 +78,13 @@ func main() {
 	}
 
 	podWatcher := reconcilers.NewPodWatcher(mgr.GetClient())
-	nsWatcher := reconcilers.NewNamespaceWatcher(mgr.GetClient())
+
+	var nsWatcher *reconcilers.NamespaceWatcher
```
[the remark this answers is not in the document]
evyatarmeged: NS watcher will probably won't use id/secret after all (again, pending decisions). I'll remove it

src/operator/controllers/intents_reconcilers/otterize_cloud/cloud_reconciler.go
evyatarmeged: Will check it out, but see comment above - will probably get removed from the watcher anyway.

src/watcher/cmd/main.go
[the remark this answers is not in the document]
orishoshan: You can have a default value with pflags too!

src/operator/api/v1alpha1/intents_types.go
orishoshan: Makes sense to have them aligned - they should represent the same object after all :)

src/watcher/cmd/main.go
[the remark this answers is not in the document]
evyatarmeged: But isn't it more logical to make something configurable in helm rather than in the code ?

PR #80, opened 2022-11-06
Test Kafka ACL creation and deletion

NetanelBollag:
### Description

Add tests for the controlling of Kafka ACL from intents objects. No documentation changes required

### References

[Kafka reconciler tests ticket in Notion](https://www.notion.so/otterize/Intents-Operator-Kafka-reconciler-tests-f80c92b2e14247a8906b9fbf48e02183)


src/operator/controllers/intents_reconcilers/kafka_acls_test.go
```diff
@@ -0,0 +1,328 @@
...
+
+	controller := gomock.NewController(s.T())
+	s.mockKafkaAdmin = kafkaaclsmocks.NewMockClusterAdmin(controller)
+
+	s.initKafkaIntentsAdmin(true)
+}
+
+func (s *KafkaACLReconcilerTestSuite) initKafkaIntentsAdmin(enableAclCreation bool) {
+	kafkaServersStore := s.setupServerStore(kafkaServiceName)
+	GetNewKafkaIntentsAdmin = func(serverConfig otterizev1alpha1.KafkaServerConfig, _ otterizev1alpha1.TLSSource, enableKafkaACLCreation bool) (kafkaacls.KafkaIntentsAdmin, error) {
```
orishoshan, line 98: Instead of having this global variable, pass the factory function to `NewKafkaACLReconciler`.

src/operator/controllers/intents_reconcilers/sarama_cluster_admin_matcher.go
```diff
@@ -0,0 +1,76 @@
...
+	return true
+}
+
+func MatchSaramaResource(s SaramaResource) gomock.Matcher {
+	return &s
+}
+
+type SaramaAclFilter sarama.AclFilter
+
+// Not used, required only to implement gomock.Matcher interface
```
orishoshan, line 55: This appears in test results. For example, when a test fails and it wants to show what's the comparison it made, it will use this `String()` to print the two sides of the comparison.

src/operator/controllers/kafkaacls/intentsAdmin.go
```diff
@@ -29,11 +29,19 @@ var (
...
+type KafkaIntentsAdmin interface {
+	ApplyServerTopicsConf(topicsConf []otterizev1alpha1.TopicConfig) error
+	ApplyClientIntents(clientName string, clientNamespace string, intents []otterizev1alpha1.Intent) error
+	RemoveClientIntents(clientName string, clientNamespace string) error
+	RemoveAllIntents() error
+	Close()
+}
+
+type KafkaIntentsAdminImpl struct {
+	KafkaServer            otterizev1alpha1.KafkaServerConfig
```
orishoshan, line 41: I looked but couldn't find a reason for all the fields to be public?
They're also implementation details, so they should be private and internal to the KafkaIntentsAdminImpl

orishoshan (changes requested):

src/operator/controllers/kafkaacls/intentsAdmin.go
NetanelBollag: It was necessary to create the object from the tests, I add `GetMockIntentsAdminFactory` function who create it inside the module to avoid making the fields public

src/operator/controllers/intents_reconcilers/sarama_cluster_admin_matcher.go
NetanelBollag: Implemented

src/operator/controllers/kafkaacls/intentsAdmin.go
```diff
@@ -600,3 +609,9 @@ func (a *KafkaIntentsAdmin) deleteConsumerGroupWildcardACLs() (int, error) {
 	}
 	return len(matchingAclRules), nil
 }
+
+func GetMockIntentsAdminFactory(clusterAdmin sarama.ClusterAdmin, usernameMapping string) IntentsAdminFactoryFunction {
```
orishoshan, line 613: This function should be part of the tests rather than here

orishoshan (approved):

PR #86, opened 2022-11-30
Adding Cloud reconciler with namespace reporting capabilities to Otterize cloud

evyatarmeged:
### Description
Cloud reconciler with report namespace capabilities.


src/operator/controllers/intents_reconcilers/otterize_cloud/cloud_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	clientIntents := otterizev1alpha1.ClientIntents{}
+	err := r.Get(ctx, req.NamespacedName, &clientIntents)
+	// In case of "Not found" k8s error, we report intents and namespace normally
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return ctrl.Result{}, err
+	}
+
+	if err = r.reportNamespace(ctx, req.Namespace); err != nil {
```
orishoshan, line 67: Utilize some sort of cache here - don't report a namespace that has already been reported successfully.
This reconciles on every ClientIntent change, so it happens many times.
I recommend keeping a `sync.Map` that holds a `sync.Once` for every namespace, then run `reportNamespace` in that sync.Once.

src/operator/controllers/intents_reconcilers/otterize_cloud/cloud_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *OtterizeCloudReconciler) reportNamespace(ctx context.Context, namespace string) error {
+	res, err := kubernetes.ReportKubernetesNamespace(ctx, r.otterizeClient, namespace)
+	if err != nil {
+		return err
+	}
+	// Return value is bool - returns true if anything was updated in the DB
+	if res.ReportKubernetesNamespace {
```
orishoshan, line 80: Don't use the return value as an indicator for success in the API - use the lack of an error for that. If the boolean is only there since you need to return something, just ignore it here (and don't vary the value in the backend)

orishoshan (approved):

src/operator/controllers/intents_reconcilers/otterize_cloud/cloud_reconciler.go
evyatarmeged: Are you sure `sync.Once` works here ? what if the call errored for some reason ? I don't think it will be called again.
I'll the `Store` and `Load` map functions, and we can change it later if necessary.

src/operator/controllers/intents_reconcilers/otterize_cloud/cloud_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func NewOtterizeCloudReconciler(
+	client client.Client,
+	scheme *runtime.Scheme,
+	otterizeClientID,
+	otterizeClientSecret,
+	cloudAddr string) *OtterizeCloudReconciler {
+
+	cfg := clientcredentials.Config{
```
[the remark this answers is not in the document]
NetanelBollag: @orishoshan @evyatarmeged I resolve this, implemented in my last commit

NetanelBollag:
@orishoshan I suggest you review my latest work, in the two last commits and approve
After that we can merge into develop, my future work on KafkaServerConfig upload will be a new PR into develop

PR #87, opened 2022-12-05
Move cloud client logic to shared module 

NetanelBollag:
### Description

Extract cloud client to a shared module so we can use it in other Otterize projects


src/operator/main.go
```diff
@@ -161,10 +160,17 @@ func main() {
 		logrus.WithError(err).Fatal("unable to init index for ingress")
 	}
 
+	graphqlClient, err = otterizecloudclient.NewClient(context.Background())
```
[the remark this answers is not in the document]
NetanelBollag: done

src/operator/main.go
```diff
@@ -161,10 +158,17 @@ func main() {
 		logrus.WithError(err).Fatal("unable to init index for ingress")
 	}
 
+	otterizeCloudClient, ok, err = otterizecloud.NewClient(context.Background())
+	if !ok {
+		logrus.Info("missing configuration for cloud integration, disabling cloud communication")
+	} else if err != nil {
```
orishoshan, line 164: Check the error even if !ok, since when an error is returned ok is still false. Basically, unchain the if,else

NetanelBollag: fixed

orishoshan (approved):

PR #89, opened 2022-12-07
Upload kafka server config to cloud

NetanelBollag:
### Description

Upload KafkaServerConfig CRD to cloud . This will allow cloud users to see services defined by their Kafka servers configuration with their Kafka specific spec even when those aren't referenced by intents or identified by mapper.

### Testing


orishoshan:
@NetanelBollag - please add the motivation for this change to the PR description

src/operator/controllers/kafkaserverconfig_controller.go
```diff
@@ -33,33 +36,43 @@ import (
...
 
-func NewKafkaServerConfigReconciler(client client.Client, scheme *runtime.Scheme, serversStore *kafkaacls.ServersStore,
-	operatorPodName string, operatorPodNameSpace string) *KafkaServerConfigReconciler {
+func NewKafkaServerConfigReconciler(
+	client client.Client,
+	scheme *runtime.Scheme,
+	serversStore kafkaacls.ServersStore,
+	operatorPodName string,
+	operatorPodNameSpace string,
+	otterizeCloudClient otterizecloud.CloudApi,
```
orishoshan, line 67: Please name this either CloudClient (I think it more accurately reflects what it is, even as an interface) or CloudAPI (following the Golang style guide - abbreviations should be all uppercase, e.g. CloudAPI not CloudApi, GraphQL not GraphQl, and so on)

src/operator/controllers/intents_reconcilers/otterizecloud/cloud_api.go
```diff
@@ -38,3 +39,14 @@ func (c *CloudApiImpl) ReportKubernetesNamespace(ctx context.Context, namespace
 	}
 	return nil
 }
+
+func (c *CloudApiImpl) ReportKafkaServerConfig(ctx context.Context, namespace string, source string, server graphqlclient.KafkaServerConfigInput) error {
+	res, err := graphqlclient.ReportKafkaServerConfig(ctx, c.client, namespace, source, server)
+	if err != nil {
+		return err
+	}
+	if res.ReportKafkaServerConfig {
```
orishoshan, line 48: What happens if it's false? Function didn't succeed but returns no error? Even though GQL returns a value, since only the error matters - please ignore the return value and only use the error.

src/operator/controllers/kafkaserverconfig_controller.go
```diff
@@ -258,10 +271,46 @@ func (r *KafkaServerConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	r.RecordNormalEvent(kafkaServerConfig, ReasonAppliedKafkaServerConfigFailed, "successfully applied server config")
 	return ctrl.Result{}, nil
 }
 
+func (r *KafkaServerConfigReconciler) uploadKafkaServerConfig(ctx context.Context, kafkaServerConfig *otterizev1alpha1.KafkaServerConfig) error {
+	if r.otterizeClient == nil {
+		return nil
+	}
+
+	err := r.otterizeClient.ReportKubernetesNamespace(ctx, kafkaServerConfig.Namespace)
```
orishoshan, line 288: I believe the other place that reports namespaces uses a cache to avoid reporting twice, see if you can use it too (maybe put it in the client?)

orishoshan (changes requested):

src/operator/controllers/intents_reconcilers/otterizecloud/cloud_api.go
NetanelBollag: Well, the way I wrote it is a mistake, but actually if `err == nil` so the boolean is true. 
We should definitely find a way to return Null

src/operator/controllers/kafkaserverconfig_controller.go
NetanelBollag: I don't want to cache here, since it's very rare to create this object, if it's reported twice it due to removal and re-apply of the resource. So we do want to report it and log the change in the access graph

orishoshan (approved):

PR #93, opened 2022-12-19
Handle Kafka intents with no topics as if they allow all topics

orishoshan:
Previously, Kafka intents with no topics would simply not allow any access. Intents that are not specific about which resources are allowed should allow all access of the kind that isn't specified. For example, a HTTP intent with no HTTP method or path specified just means that you can access using HTTP, to any path and any method.

Similarly, a Kafka intent with no topics or operations specified should create an ACL that allows access to all topics and operations, and a Kafka intent with only the topic name specified should create an ACL that allows all operations for that topic.

### DIFF
--- a/src/shared/testbase/testsuitebase.go
+++ b/src/shared/testbase/testsuitebase.go
@@ -29,6 +29,7 @@ import (
 const waitForCreationInterval = 20 * time.Millisecond
 const waitForCreationTimeout = 10 * time.Second
 const waitForDeletionTSTimeout = 3 * time.Second
+const maxNamespaceLength = 63
 
 type ControllerManagerTestSuiteBase struct {
 	suite.Suite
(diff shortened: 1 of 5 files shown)


src/shared/testbase/testsuitebase.go
```diff
@@ -61,7 +62,7 @@ func (s *ControllerManagerTestSuiteBase) BeforeTest(_, testName string) {
 		s.Require().NoError(err)
 	}()
 
-	s.TestNamespace = strings.ToLower(fmt.Sprintf("%s-%s", testName, time.Now().Format("20060102150405")))
+	s.TestNamespace = strings.ToLower(fmt.Sprintf("%s-%s", testName, time.Now().Format("20060102150405")))[:maxNamespaceLength]
```
orishoshan, line 65: Bugfix - if test name was too long, then the resulting namespace was >63 chars, which made the test fail to initialize.

omris94 (approved):

PR #94, opened 2022-12-21
Adding intents reporting to Otterize cloud in the cloud reconciler 

evyatarmeged:
### Description
Added sending intents to Otterize cloud as part of the cloud reconciler.
Small additional tweaks to cloud client. Added `sync.Map` to store reported namespaces in the cloud reconciler.


src/shared/otterizecloudclient/config.go
```diff
@@ -9,7 +9,7 @@ const (
 	ApiClientIdKey            = "client-id"
 	ApiClientSecretKey        = "client-secret"
 	OtterizeAPIAddressKey     = "api-address"
-	OtterizeAPIAddressDefault = "https://app.otterize.com/api"
+	OtterizeAPIAddressDefault = "https://app.otterize.com"
```
orishoshan, line 12: How does this change make sense? Don't all the API endpoints have /api?

orishoshan (approved):

PR #95, opened 2022-12-26
Report kafka config without source

NetanelBollag:
### Description

Current implementation generate the source field in the backend, so the operator shouldn't report it


orishoshan:
Just to clarify - the source field that was previously there is an implementation detail left over from an old API, and as the new API does not have it, it was removed here.

src/operator/controllers/kafkaserverconfig_controller_test.go
```diff
@@ -134,6 +135,10 @@ func (s *KafkaServerConfigReconcilerTestSuite) reconcile(namespacedName types.Na
 		res, err = s.reconciler.Reconcile(context.Background(), ctrl.Request{
 			NamespacedName: namespacedName,
 		})
+
+		if err != nil && strings.Contains(err.Error(), "the object has been modified; please apply your changes to the latest version and try again") {
```
orishoshan, line 139: use `if k8serrors.IsConflict` instead - see examples elsewhere

orishoshan (approved):
fix comment then merge

src/operator/controllers/kafkaserverconfig_controller_test.go
NetanelBollag: Did it

orishoshan (approved):

PR #101, opened 2023-01-02
Added enable-enforcement=false flag that disables enforcement globally

amit7itz:
This PR adds the --enable-enforcement=false flag, which can be used in order to disable all the enforcement abilities of the intents operator at once.
If enable-enforcement is true (which is the default), the other flags can be used in order to disable only specific enforcement types (e.g. disable only Kafka ACLs)

Tests:
Added unit tests to verify network policies and Kafka ACLs are not created when enable-enforcement=false

### DIFF
--- a/src/operator/controllers/kafkaserverconfig_controller_test.go
+++ b/src/operator/controllers/kafkaserverconfig_controller_test.go
@@ -99,7 +99,7 @@ func (s *KafkaServerConfigReconcilerTestSuite) BeforeTest(_, testName string) {
 }
 
 func getMockIntentsAdminFactory(mockIntentsAdmin *kafkaaclsmocks.MockKafkaIntentsAdmin) kafkaacls.IntentsAdminFactoryFunction {
-	return func(kafkaServer otterizev1alpha1.KafkaServerConfig, _ otterizev1alpha1.TLSSource, enableKafkaACLCreation bool, globalEnforceSetting bool) (kafkaacls.KafkaIntentsAdmin, error) {
+	return func(kafkaServer otterizev1alpha1.KafkaServerConfig, _ otterizev1alpha1.TLSSource, enableKafkaACLCreation bool, enforcementEnabledGlobally bool) (kafkaacls.KafkaIntentsAdmin, error) {
 		return mockIntentsAdmin, nil
 	}
 }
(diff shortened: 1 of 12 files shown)


src/operator/controllers/kafkaserverconfig_controller_test.go
```diff
@@ -133,6 +134,9 @@ func (s *KafkaServerConfigReconcilerTestSuite) reconcile(namespacedName types.Na
 		res, err = s.reconciler.Reconcile(context.Background(), ctrl.Request{
 			NamespacedName: namespacedName,
 		})
+		if k8serrors.IsConflict(err) {
+			res.Requeue = true
```
amit7itz, line 138: This is a bugfix for a flaky test, solution by @NetanelBollag

src/operator/controllers/kafkaserverconfig_controller_test.go
```diff
@@ -163,16 +167,12 @@ func (s *KafkaServerConfigReconcilerTestSuite) TestReUploadKafkaServerConfigOnFa
 	kafkaServerConfig.SetNamespace(s.TestNamespace)
 	s.AddKafkaServerConfig(&kafkaServerConfig)
 
-	// Set go mock expectations for failure
-	s.mockIntentsAdmin.EXPECT().ApplyServerTopicsConf(kafkaServerConfig.Spec.Topics).Return(nil).Times(1)
-	s.mockIntentsAdmin.EXPECT().Close()
```
amit7itz, line 168: This is also a fix to a test by @NetanelBollag. the logic here changed because of the namespaces PR.

src/operator/main.go
```diff
@@ -221,6 +226,10 @@ func main() {
 		logrus.WithError(err).Fatal("unable to set up ready check")
 	}
 
+	if !enforcementEnabledGlobally {
+		logrus.Info("Running with %s=false, won't perform any enforcement", enableEnforcementKey)
```
amitlicht, line 230: @amit7itz the linter is right :) use Infof

src/operator/main.go
```diff
@@ -135,9 +140,9 @@ func main() {
 		logrus.WithError(err).Fatal(err, "unable to start manager")
 	}
 
-	kafkaServersStore := kafkaacls.NewServersStore(tlsSource, enableKafkaACLCreation, kafkaacls.NewKafkaIntentsAdmin)
+	kafkaServersStore := kafkaacls.NewServersStore(tlsSource, enableKafkaACLCreation, kafkaacls.NewKafkaIntentsAdmin, enforcementEnabledGlobally)
```
amitlicht, line 143: This is a matter of taste and simplicity, so putting this here as a thought and let's discuss if you disagree:

IMO it would have been simpler (and clear enough) to keep the `enforcementEnabledGlobally` flag only here on main (and probably in the component that's going to report this status upwards to the cloud), and have the specific reconcilers use their dedicated flags without having to pass the `enforcementEnabledGlobally` around. Meaning: 
```
kafkaServersStore := kafkaacls.NewServersStore(tlsSource, enableKafkaACLCreation && enforcementEnabledGlobally, kafkaacls.NewKafkaIntentsAdmin)
...
endpointReconciler := external_traffic.NewEndpointsReconciler(mgr.GetClient(), mgr.GetScheme(), enforcementEnabledGlobally && autoCreateNetworkPoliciesForExternalTraffic)
... 
```
Another advantage of this approach is that it's safer, in that you don't have to make sure you propagate and support this global flag in all existing and future reconcilers. Every reconcile should just have an on/off switch, and the global disable flag just disables all switches.

amitlicht (changes requested):

src/operator/main.go
amit7itz: nice 😬

amitlicht (approved):

PR #103, opened 2023-01-03
Fixed error when running without cloud connectivity

amit7itz:
This PR fixed an error when running without cloud credentials
```
missing cloud integration client ID
```


NetanelBollag (approved):
So great!

PR #108, opened 2023-01-08
Fix report intents to cloud

NetanelBollag:
Reporting applied intents to cloud fail since the cloud expect intents type whenever intents body object exists, and until this PR body object is sent empty instead of nil.

This PR enable sending nil objects and also add cloud reconciler test infrastructure

### DIFF
--- a/src/operator/api/v1alpha1/intents_types.go
+++ b/src/operator/api/v1alpha1/intents_types.go
@@ -195,11 +195,12 @@ func (in *Intent) typeAsGQLType() graphqlclient.IntentType {
 	}
 }
 
-func (in *ClientIntentsList) FormatAsOtterizeIntents(intentsObjNamespace string) ([]graphqlclient.IntentInput, error) {
+func (in *ClientIntentsList) FormatAsOtterizeIntents() ([]*graphqlclient.IntentInput, error) {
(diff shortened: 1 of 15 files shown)


src/operator/api/v1alpha1/intents_types.go
```diff
@@ -195,35 +195,43 @@ func (in *Intent) typeAsGQLType() graphqlclient.IntentType {
...
 
 	intentInput := graphqlclient.IntentInput{
-		ClientName:      clientName,
-		ServerName:      in.Name,
-		Namespace:       resourceNamespace,
-		ServerNamespace: in.ResolveServerNamespace(resourceNamespace),
+		ClientName:      lo.ToPtr(clientName),
+		ServerName:      lo.ToPtr(in.Name),
+		Namespace:       lo.ToPtr(resourceNamespace),
+		ServerNamespace: toPtrOrNil(in.ResolveServerNamespace(resourceNamespace)),
```
evyatarmeged, line 230: Are we sure this won't break the cloud side of the reporting ? How can `ClientName` be optional ? 
I understand why `ServerName` can be null, but `ClientName` and `Namespace` (which can always be derived form the resource namespace) should always be present, and i think the backend expects it to be so.

evyatarmeged (approved):

Issue #90, opened 2022-12-09
Change namespace in intents file from a property to a suffix on the name


usarid:
In K8s client intents files, the client indicates which service it intends to call by specifying the service's `name`; the assumption is that the target service is in the same namespace as the client. To target a service in a different namespace (of the same cluster), the client must specify the target's namespace. That is currently expressed as:
```
calls:
  - name: target-service
  - namespace: target-namespace
```
This issue proposes to keep the semantics but change the format to:
```
calls:
  - name: target-service.target-namespace
```
The new format is more aligned with how calling services is handled within and outside Kubernetes, in that it follows a DNS pattern. 

This would be a breaking change, but this early in the life of this project I feel we can still make changes like this. Holler if you disagree.

evyatarmeged:
This was just merged in. Feel free to reopen if required.

PR #114, opened 2023-01-16
Remove namespace from client intents format, bump version to v1alpha2

evyatarmeged:
### Description
`clientintents` custom resource will no longer support the `namespace` field when describing the client's call list. 
Instead, users should use the new format `serverName.namespace` to specify servers that aren't in the same namespace as the client. 

### References
Solves #90 

### Checklist


src/operator/Makefile
```diff
@@ -145,3 +145,6 @@ envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
 	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
+.PHONY: love
```
orishoshan, line 148: ?

orishoshan (changes requested):

orishoshan (approved):

PR #117, opened 2023-01-23
Convert Kafka operation enum to cloud format

NetanelBollag:
The Kafka operation didn't have structured conversion from k8s format to cloud format, just pure string conversation. There was also an missing enum value for Kafka operation. This all caused error when uploading Kafka operation to cloud.
This PR implement the missing value.

### DIFF
--- a/src/operator/api/v1alpha2/intents_types.go
+++ b/src/operator/api/v1alpha2/intents_types.go
@@ -226,12 +226,42 @@ func toPtrOrNil(s string) *string {
 	return lo.ToPtr(s)
 }
 
+func kafkaOperationK8sToCloud(op KafkaOperation) graphqlclient.KafkaOperation {
+	switch op {
+	case KafkaOperationAll:
+		return graphqlclient.KafkaOperationAll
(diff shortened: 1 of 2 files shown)


src/operator/api/v1alpha2/intents_types.go
```diff
@@ -226,12 +226,42 @@ func toPtrOrNil(s string) *string {
...
+	case KafkaOperationClusterAction:
+		return graphqlclient.KafkaOperationClusterAction
+	case KafkaOperationDescribeConfigs:
+		return graphqlclient.KafkaOperationDescribeConfigs
+	case KafkaOperationAlterConfigs:
+		return graphqlclient.KafkaOperationAlterConfigs
+	case KafkaOperationIdempotentWrite:
+		return graphqlclient.KafkaOperationIdempotentWrite
+	default:
+		panic(fmt.Sprintf("Unknown KafkaOperation: %s", op))
```
orishoshan, line 254: do logrus.Panic so that it logs the line number, etc

orishoshan (approved):

PR #104, opened 2023-01-05
Reporting intents operator configuration to the cloud

amit7itz:
This PR makes the intents operator report its configuration to the cloud at start

PR #123, opened 2023-01-25
Report network policy & Kafka ACL enforcement state in addition to global enforcement state

orishoshan:
### Description

A previous change #104 added the ability to report global enforcement state to Otterize Cloud - this PR adds the ability to report network policy and Kafka ACL enforcement state individually, in addition to the global state.

### DIFF
--- a/src/operator/controllers/intents_reconcilers/otterizecloud/cloud_api.go
+++ b/src/operator/controllers/intents_reconcilers/otterizecloud/cloud_api.go
@@ -11,7 +11,7 @@ import (
 type CloudClient interface {
 	ReportKafkaServerConfig(ctx context.Context, namespace string, servers []graphqlclient.KafkaServerConfigInput) error
 	ReportAppliedIntents(ctx context.Context, namespace *string, intents []*graphqlclient.IntentInput) error
-	ReportIntentsOperatorConfiguration(ctx context.Context, enableEnforcement bool) error
+	ReportIntentsOperatorConfiguration(ctx context.Context, globalEnforcementEnabled bool, networkPolicyEnforcementEnabled bool, kafkaACLEnforcementEnabled bool) error
 	ReportComponentStatus(ctx context.Context, component graphqlclient.ComponentType)
 }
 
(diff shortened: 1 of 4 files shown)


src/operator/main.go
```diff
@@ -230,20 +217,24 @@ func main() {
...
 
-	ctx, cancelFunc := context.WithTimeout(context.Background(), 10*time.Second)
+	signalHandlerCtx := ctrl.SetupSignalHandler()
+	timeoutCtx, cancelFunc := context.WithTimeout(signalHandlerCtx, 10*time.Second)
 	defer cancelFunc()
-	cloudClient, connectedToCloud, err := otterizecloud.NewClient(ctx)
+	otterizeCloudClient, connectedToCloud, err := otterizecloud.NewClient(timeoutCtx)
 	if err != nil {
-		logrus.WithError(err).Warning("Failed to connect to Otterize cloud")
+		logrus.WithError(err).Fatal("Failed to initialize Otterize Cloud client")
```
amit7itz, line 225: IMO the operator shouldn't crash if it cannot connect to the cloud. Operator downtime means potential access issues between the services in the cluster, so it should be as resistant as possible to cloud connectivity loss.

src/operator/main.go
```diff
@@ -230,20 +217,24 @@ func main() {
...
 	}
 	if connectedToCloud {
-		err := cloudClient.ReportIntentsOperatorConfiguration(ctx, enforcementEnabledGlobally)
+		err := otterizeCloudClient.ReportIntentsOperatorConfiguration(timeoutCtx, enforcementEnabledGlobally, enableNetworkPolicyCreation, enableKafkaACLCreation)
 		if err != nil {
-			logrus.WithError(err).Warning("Failed to report configuration to the cloud")
+			logrus.WithError(err).Fatal("Failed to report configuration to the cloud")
```
amit7itz, line 230: same here

src/operator/controllers/intents_reconcilers/otterizecloud/cloud_api.go
```diff
@@ -57,9 +57,13 @@ func (c *CloudClientImpl) ReportAppliedIntents(
 
 func (c *CloudClientImpl) ReportIntentsOperatorConfiguration(
 	ctx context.Context,
-	enableEnforcement bool) error {
+	globalEnforcementEnabled bool,
+	networkPolicyEnforcementEnabled bool,
+	kafkaACLEnforcementEnabled bool) error {
```
amit7itz, line 62: Please make the function receive the IntentsOperatorConfigurationInput instead, it reduces the chances that people accidentally pass the parameters in the wrong order.

amit7itz (approved):
See comments

PR #127, opened 2023-02-05
Fix small timeout in token acquisition

NetanelBollag:
### Description

Until this PR token source timeout received context with timeout of 10sec from application start.
Since token acquisition happen after 24 hours, it immediately gets a timeout and exit.

This PR create token source with timeout on httpclient, like the package expects the user to do

### DIFF
--- a/src/operator/controllers/intents_reconcilers/otterizecloud/cloud_reconciler.go
+++ b/src/operator/controllers/intents_reconcilers/otterizecloud/cloud_reconciler.go
@@ -5,6 +5,7 @@ import (
 	otterizev1alpha2 "github.com/otterize/intents-operator/src/operator/api/v1alpha2"
 	"github.com/otterize/intents-operator/src/shared/injectablerecorder"
 	"github.com/samber/lo"
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -43,8 +44,11 @@ func (r *OtterizeCloudReconciler) Reconcile(ctx context.Context, req reconcile.R
 	}
 
 	if err = r.otterizeClient.ReportAppliedIntents(ctx, lo.ToPtr(req.Namespace), intentsInput); err != nil {
+		logrus.WithError(err).Error("failed to report applied intents")
 		return ctrl.Result{}, err
 	}
+
+	logrus.Infof("successfully reported applied intents %d", len(clientIntentsList.Items))
 
 	return ctrl.Result{}, nil
 }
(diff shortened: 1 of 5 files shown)


src/operator/controllers/intents_reconcilers/otterizecloud/cloud_reconciler.go
```diff
@@ -43,8 +44,11 @@ func (r *OtterizeCloudReconciler) Reconcile(ctx context.Context, req reconcile.R
 	}
 
 	if err = r.otterizeClient.ReportAppliedIntents(ctx, lo.ToPtr(req.Namespace), intentsInput); err != nil {
+		logrus.WithError(err).Error("failed to report applied intents")
 		return ctrl.Result{}, err
 	}
 
+	logrus.Infof("successfully reported applied intents %d", len(clientIntentsList.Items))
```
amit7itz, line 51: move the %d to the middle: "successfully reported %d applied intents"

src/shared/otterizecloud/otterizecloudclient/cloud_client.go
```diff
@@ -31,9 +33,13 @@ func NewClient(ctx context.Context) (graphql.Client, bool, error) {
...
-	graphqlUrl := fmt.Sprintf("%s/graphql/v1beta", apiAddress)
-	httpClient := oauth2.NewClient(ctx, tokenSrc)
+	// Timeout for oauth token acquisition is set using the http client passed to the token source context
+	clientWithTimeout := &http.Client{Timeout: clientTimeout}
+	ctxWithClient := context.WithValue(ctx, oauth2.HTTPClient, clientWithTimeout)
 
+	tokenSrc := cfg.TokenSource(ctxWithClient)
+	graphqlUrl := fmt.Sprintf("%s/graphql/v1beta", apiAddress)
+	httpClient := oauth2.NewClient(ctxWithClient, tokenSrc)
+	httpClient.Timeout = clientTimeout
```
amit7itz, line 43: Sure this is required? doesn't it use the client you provided?

src/shared/otterizecloud/otterizecloudclient/cloud_client.go
```diff
@@ -31,9 +33,13 @@ func NewClient(ctx context.Context) (graphql.Client, bool, error) {
 		AuthStyle:    oauth2.AuthStyleInParams,
 	}
 
-	tokenSrc := cfg.TokenSource(ctx)
-	graphqlUrl := fmt.Sprintf("%s/graphql/v1beta", apiAddress)
-	httpClient := oauth2.NewClient(ctx, tokenSrc)
+	// Timeout for oauth token acquisition is set using the http client passed to the token source context
```
amit7itz, line 36: Right now it might sound like this timeout is only relevant to oauth, make it clearer that this timeout is for all HTTP calls (including oauth)

amit7itz (approved):

src/shared/otterizecloud/otterizecloudclient/cloud_client.go
NetanelBollag: Yes, they right it in the documentation, I'll add comment

PR #145, opened 2023-03-27
Add mock tests istio policy

NetanelBollag:
Add test to istio policy creator


src/shared/istiopolicy/creator.go
```diff
@@ -137,11 +137,10 @@ func (c *Creator) updateOrCreatePolicy(
...
-	if existingPolicy.Spec.Selector == nil || newPolicy.Spec.Selector == nil {
-		return false, fmt.Errorf("policy pod selector is nil")
-	}
-
-	if len(existingPolicy.Spec.Rules) == 0 || len(existingPolicy.Spec.Rules[0].From) == 0 {
+	if len(existingPolicy.Spec.Rules) == 0 ||
+		len(existingPolicy.Spec.Rules[0].From) == 0 ||
+		existingPolicy.Spec.Selector == nil ||
+		len(existingPolicy.Spec.Selector.MatchLabels) == 0 {
 		logrus.Warning("found existing policy with bad format, overwriting")
```
orishoshan, line 144: When does this happen? What does bad format mean? If it is possible to create a bad policy, we should change that flow so it's no longer possible. I think all flows should either end in success or error, logging a warning probably means something we would rather not happen, happened.

orishoshan (changes requested):
see comment

src/shared/istiopolicy/creator.go
NetanelBollag: If it happens it's not because of us, but a result of some manually created policy. 
If there are no Rules we would crush due to nil reference, and like any output we receive from outer-space, it can't kill us.

NetanelBollag: Ok, I'll change that

PR #146, opened 2023-03-27
Pod watcher ignore policy creation when missing SA or deleted intents

NetanelBollag:
Pod watcher shouldn't create Istio authorization policies when the pod is missing Service Account or intents are being deleted

### DIFF
--- a/src/watcher/reconcilers/pods.go
+++ b/src/watcher/reconcilers/pods.go
@@ -23,9 +23,7 @@ import (
 )
 
 const (
-	OtterizeClientNameIndexField    = "spec.service.name"
-	ReasonPodMissingServiceAccount  = "PodMissingServiceAccount"
-	ReasonRemovingIstioPolicyFailed = "RemovingIstioPolicyFailed"
+	OtterizeClientNameIndexField = "spec.service.name"
 )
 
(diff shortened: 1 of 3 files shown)


src/watcher/reconcilers/pods.go
```diff
@@ -127,16 +137,28 @@ func (p *PodWatcher) istioEnforcementEnabled() bool {
 	return viper.GetBool(operatorconfig.IstioFeatureFlagEnabledKey) && viper.GetBool(operatorconfig.EnableIstioPolicyKey)
 }
 
-func (p *PodWatcher) createIstioPolicies(ctx context.Context, intents otterizev1alpha2.ClientIntents, pod v1.Pod) {
+func (p *PodWatcher) createIstioPolicies(ctx context.Context, intents otterizev1alpha2.ClientIntents, pod v1.Pod) error {
 	serviceAccountName := pod.Spec.ServiceAccountName
 	if serviceAccountName == "" {
```
orishoshan, line 142: I think this case cannot happen as pods must always have service accounts

orishoshan (changes requested):
see comment

orishoshan (approved):

PR #148, opened 2023-04-02
Manage Istio authorization policies by label instead of name

NetanelBollag:
This PR makes it easier to manage authorization by policy labels instead of by policy name, allowing to deletion of outdated policies. It first reads all policies created for the client, updates the existing ones, creates new and deletes the rest.


src/shared/istiopolicy/creator.go
```diff
@@ -42,71 +45,110 @@ func NewCreator(client client.Client, recorder *injectablerecorder.InjectableRec
...
+		client.MatchingLabels{v1alpha2.OtterizeIstioClientAnnotationKey: clientFormattedIdentity})
+	if err != nil {
+		c.recorder.RecordWarningEventf(clientIntents, ReasonGettingIstioPolicyFailed, "could not get istio policies: %s", err.Error())
+		return err
+	}
 
-			logrus.WithError(err).Errorf("could not get istio policy %s", policyName)
-			return err
-		}
+	validPolicies, err := c.handelPolicies(ctx, clientIntents, clientNamespace, clientServiceAccount, existingPolicies)
```
orishoshan, line 82: handel -> handle

src/shared/istiopolicy/creator.go
```diff
@@ -42,71 +45,110 @@ func NewCreator(client client.Client, recorder *injectablerecorder.InjectableRec
...
+		newPolicy := c.generateAuthorizationPolicy(clientIntents, intent, clientNamespace, clientServiceAccount)
+		existingPolicy, found := c.findPolicy(existingPolicies, newPolicy)
+		if found {
+			err := c.UpdatePolicy(ctx, existingPolicy, newPolicy)
+			if err != nil {
+				c.recorder.RecordWarningEventf(clientIntents, ReasonUpdatingIstioPolicyFailed, "failed to update istio policy: %s", err.Error())
+				return goset.Set[types.UID]{}, err
+			}
+			validPolicies.Add(existingPolicy.UID)
+		} else {
```
orishoshan, line 122: Turn this into `if !found { ... return }` then the above block

src/shared/istiopolicy/creator.go
```diff
@@ -42,71 +45,110 @@ func NewCreator(client client.Client, recorder *injectablerecorder.InjectableRec
...
 
-func (c *Creator) handleAuthorizationPolicy(
-	ctx context.Context,
-	intents *v1alpha2.ClientIntents,
-	clientIntentsNamespace string,
-	clientServiceAccountName string,
-) error {
-	for _, intent := range intents.GetCallsList() {
-		if c.namespaceNotAllowed(intent, clientIntentsNamespace) {
+func (c *Creator) handelPolicies(ctx context.Context, clientIntents *v1alpha2.ClientIntents, clientNamespace string, clientServiceAccount string, existingPolicies v1beta1.AuthorizationPolicyList) (goset.Set[types.UID], error) {
```
orishoshan, line 100: Rename to createOrUpdatePolicies to make it clear that this doesn't do all the handling

src/shared/istiopolicy/creator.go
orishoshan: Also let's  call `validPolicies` the `newOrUpdatedPolicies` or something that indicates that they're not valid or invalid, but just new or updated.

src/watcher.Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN go mod download
 COPY . .
 
 RUN go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
-RUN KUBEBUILDER_ASSETS=`setup-envtest use 1.24.1 -p path` go test ./watcher/...
+#RUN KUBEBUILDER_ASSETS=`setup-envtest use 1.24.1 -p path` go test ./watcher/...
```
orishoshan, line 12: ?

orishoshan (approved):

PR #149, opened 2023-04-03
Add HTTP enforcement

NetanelBollag:
Support HTTP resource protection using Istio Authorization Policies

### Testing

### DIFF
--- a/src/shared/istiopolicy/creator.go
+++ b/src/shared/istiopolicy/creator.go
@@ -16,6 +16,7 @@ import (
 	"istio.io/client-go/pkg/apis/security/v1beta1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


src/shared/istiopolicy/creator.go
```diff
@@ -194,8 +195,35 @@ func (c *Creator) getPolicyName(intents *v1alpha2.ClientIntents, intent v1alpha2
...
+func compareHTTPRules(existingRules []*v1beta12.Rule_To, newRules []*v1beta12.Rule_To) bool {
+	if len(existingRules) == 0 && len(newRules) == 0 {
+		return true
+	}
+
+	if len(existingRules) != len(newRules) {
+		return false
+	}
+
+	for i := range existingRules {
```
orishoshan, line 212: If you're relying on the order for comparison, you should probably sort them, otherwise you risk an equal set of rules not actually being equal.
Or, place them in a map for comparison.

src/shared/istiopolicy/creator_test.go
```diff
@@ -104,6 +104,359 @@ func (s *CreatorTestSuite) TestCreate() {
...
+			Rules: []*v1beta12.Rule{
+				{
+					To: []*v1beta12.Rule_To{
+						{
+							Operation: &v1beta12.Operation{
+								Paths: []string{
+									"/login",
+								},
+								Methods: []string{
+									"GET",
```
orishoshan, line 172: Consider using constants from `http`: `http.MethodGet`

src/shared/istiopolicy/creator.go
```diff
@@ -246,3 +286,27 @@ func (c *Creator) generateAuthorizationPolicy(
...
+	}
+
+	return operations
+}
+
+func (c *Creator) intentsMethodsToIstioMethods(intent []v1alpha2.HTTPMethod) []string {
+	istioMethods := make([]string, 0, len(intent))
+	for _, method := range intent {
+		// Istio documentation specifies "A list of methods as specified in the HTTP request" in uppercase
+		strMethod := strings.ToUpper(string(method))
```
orishoshan, line 307: You can change the definition of HTTPMethod so that instead of having hardcoded constants, it refers to http.MethodGet, etc.
But right now I see it's already uppercase, so you don't need to worry about it - it will be uppercase anyway

orishoshan (approved):

PR #151, opened 2023-04-03
Create intents status

NetanelBollag:
Create status for ClientIntents CRD and save shared service account info in it.


src/shared/istiopolicy/creator.go
```diff
@@ -98,6 +102,80 @@ func (c *Creator) Create(
...
+	}
+
+	logrus.Infof("Labeling intent %s with service account name %s", clientIntents.Name, clientServiceAccount)
+
+	c.recorder.RecordNormalEventf(clientIntents, ReasonServiceAccountFound, "Client intents service account found %s", clientServiceAccount)
+	return nil
+}
+
+func (c *Creator) updateStatusSharedServiceAccount(ctx context.Context, namespace string, clientServiceAccountName string) error {
+	var ClientIntentsList v1alpha2.ClientIntentsList
```
orishoshan, line 144: Please keep the existing conventions - use a lowercase first letter for the variable name, as in all the rest. This also goes for HasSharedServiceAccount

src/shared/istiopolicy/creator.go
```diff
@@ -98,6 +102,80 @@ func (c *Creator) Create(
...
+	err := c.client.List(ctx, &ClientIntentsList, client.InNamespace(namespace), client.MatchingLabels{
+		v1alpha2.OtterizeClientServiceAccountLabel: clientServiceAccountName,
```
orishoshan, line 146: Would this work with all possible values?
Kubernetes resource names are limited to 253 characters
Label values are limited to 63 characters
So please use an annotation

src/operator/config/crd/k8s.otterize.com_clientintents.yaml
```diff
@@ -111,6 +111,9 @@ spec:
             type: object
           status:
             description: IntentsStatus defines the observed state of ClientIntents
+            properties:
+              hasSharedServiceAccount:
```
orishoshan, line 115: A few notes here:
1. The status field is supposed to be user-facing, while this is used internally.
2. We discussed using the status field to reflect this information, and if we do this, we need to also reflect the other kinds of statuses here. This PR doesn't do that, so this attribute is the odd one out (everything is reflected through events)
3. Adding this to the schema will make it harder to change later, as it requires updating the CRD.

As this might be a temporary solution, let's please use an annotation instead, which keeps it as a pure implementation detail - we could change how this is implemented in the intents operator without any user-facing effect.

src/shared/istiopolicy/creator.go
```diff
@@ -98,6 +102,80 @@ func (c *Creator) Create(
...
+		}
+
+		updatedIntents.Status.HasSharedServiceAccount = HasSharedServiceAccount
+		err = c.client.Status().Patch(ctx, updatedIntents, client.MergeFrom(&intents))
+		if err != nil {
+			return err
+		}
+
+		if HasSharedServiceAccount {
+			c.recorder.RecordWarningEventf(updatedIntents, ReasonSharedServiceAccount, "Service account %s is shared by multiple clients", clientServiceAccountName)
```
orishoshan, line 172: What is the outcome? The message should say that this means that other services will also be able to gain access because of this intent, and it should also say which ones (based on the list of client intents)

src/shared/istiopolicy/creator.go
```diff
@@ -98,6 +102,110 @@ func (c *Creator) Create(
...
+
+func (c *Creator) updateServiceAccountSharedStatus(ctx context.Context, clientIntents []v1alpha2.ClientIntents, serviceAccount string) error {
+	logrus.Infof("Found %d intents with service account %s", len(clientIntents), serviceAccount)
+	isServiceAccountShared := len(clientIntents) > 1
+	sharedAccountValue := lo.Ternary(isServiceAccountShared, "true", "false")
+
+	clients := lo.Map(clientIntents, func(intents v1alpha2.ClientIntents, _ int) string {
+		return intents.Spec.Service.Name
+	})
+	clientsNames := strings.Join(clients, ",")
```
orishoshan, line 173: Can you make it ", "?

src/shared/istiopolicy/creator.go
```diff
@@ -98,6 +102,110 @@ func (c *Creator) Create(
...
+		updatedIntents := intents.DeepCopy()
+
+		updatedIntents.Annotations[v1alpha2.OtterizeSharedServiceAccountAnnotation] = sharedAccountValue
+		err := c.client.Patch(ctx, updatedIntents, client.MergeFrom(&intents))
+		if err != nil {
+			return err
+		}
+
+		if isServiceAccountShared {
+			c.recorder.RecordWarningEventf(updatedIntents, ReasonSharedServiceAccount, "Service account %s is shared and will grant access to the following clients %s", serviceAccount, clientsNames)
```
orishoshan, line 189: Can you make it: "...and will *also* grant access to the following clients*:*"
(differences: 'also' and ':')

orishoshan (approved):

PR #152, opened 2023-04-09
Upload intents status and HTTP resources to cloud

NetanelBollag:


src/operator/api/v1alpha2/intents_types.go
```diff
@@ -217,12 +217,34 @@ func (in *ClientIntentsList) FormatAsOtterizeIntents() ([]*graphqlclient.IntentI
...
+func clientIntentsStatusToCloudFormat(clientIntents ClientIntents) *graphqlclient.IntentStatusInput {
+	var status graphqlclient.IntentStatusInput
+
+	serviceAccountName, ok := clientIntents.Annotations[OtterizeClientServiceAccountAnnotation]
+	if !ok {
+		return nil
+	}
+
+	status.ServiceAccountName = toPtrOrNil(serviceAccountName)
+	value, ok := clientIntents.Annotations[OtterizeSharedServiceAccountAnnotation]
```
orishoshan, line 238: You can shorten this block by just doing if ok { lo.ToPtr }, since nil is the default value

orishoshan (approved):

src/operator/api/v1alpha2/intents_types.go
NetanelBollag: I like it to be explicit and in the same format of `if !ok {` same as line above

PR #155, opened 2023-04-10
Report missing sidecar

NetanelBollag:
### Description

Report when pod is missing Istio sidecar to intents status, and ignore its policies. 

### DIFF
--- a/src/shared/istiopolicy/creator.go
+++ b/src/shared/istiopolicy/creator.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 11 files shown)


src/shared/istiopolicy/creator.go
```diff
@@ -140,6 +148,33 @@ func (c *Creator) saveServiceAccountName(ctx context.Context, clientIntents *v1a
 	return nil
 }
 
+func (c *Creator) saveSideCarStatus(ctx context.Context, clientIntents *v1alpha2.ClientIntents, missingSideCar bool) error {
+	oldValue, ok := clientIntents.Annotations[v1alpha2.OtterizeMissingSidecarAnnotation]
```
orishoshan, line 152: For future improvement, maybe you could use the presence of the annotation as a marker, then you wouldn't have to do the string/bool parsing?

src/shared/istiopolicy/creator.go
```diff
@@ -140,6 +148,33 @@ func (c *Creator) saveServiceAccountName(ctx context.Context, clientIntents *v1a
...
+	updatedIntents.Annotations[v1alpha2.OtterizeMissingSidecarAnnotation] = strconv.FormatBool(missingSideCar)
+	err := c.client.Patch(ctx, updatedIntents, client.MergeFrom(clientIntents))
+	if err != nil {
+		logrus.WithError(err).Errorln("Failed updating intent with sidecar status")
+		c.recorder.RecordWarningEventf(clientIntents, ReasonIntentsLabelFailed, "Failed updating intent with service account name: %s", err.Error())
+		return err
+	}
+
+	if missingSideCar {
+		logrus.Infof("updating intent %s with side %v", clientIntents.Name, missingSideCar)
```
orishoshan, line 171: Probably no need for the log as you are already sending it as an event on the resource, and the user will be able to see it there. IMO it's better to keep the operator log cleaner so that the user will have an easier time seeing errors that require attention, if there are any.

orishoshan (approved):

PR #166, opened 2023-04-17
Auto-create a Kafka-typed intent for the intents operator that grants access to creating and reading ACLs

orishoshan:


src/operator/controllers/kafkaserverconfig_controller.go
```diff
@@ -200,10 +200,15 @@ func (r *KafkaServerConfigReconciler) createIntentsFromOperatorToKafkaServer(ctx
 				Name: annotatedServiceName,
 			},
 			Calls: []otterizev1alpha2.Intent{{
-				// HTTP is used here to indicate that this should only apply network policies.
-				// In the future, should be updated as declaring type HTTP becomes unnecessary.
-				Type: otterizev1alpha2.IntentTypeHTTP,
```
evyatarmeged, line 205: Won't it need untyped (netpol) as well ?

orishoshan: It doesn't, because all intents do the same as untyped intents

evyatarmeged (approved):

PR #167, opened 2023-04-18
Auto-detect Istio and don't attempt to create AuthorizationPolicy if Istio is not found

orishoshan:


src/shared/operatorconfig/config.go
```diff
@@ -10,9 +10,9 @@ import (
 
 const (
 	MetricsAddrKey                                     = "metrics-bind-address" // The address the metric endpoint binds to
-	MetricsAddrDefault                                 = ":8080"
+	MetricsAddrDefault                                 = ":8180"
```
evyatarmeged, line 13: change back

src/shared/operatorconfig/config.go
```diff
@@ -10,9 +10,9 @@ import (
...
 	ProbeAddrKey                                       = "health-probe-bind-address" // The address the probe endpoint binds to
-	ProbeAddrDefault                                   = ":8081"
+	ProbeAddrDefault                                   = ":8181"
```
evyatarmeged, line 15: change back

PR #182, opened 2023-05-17
Remove redundant call to k8s api

NetanelBollag:
The code removed call client.Get for operator pod, but nothing is done we the pod read. The only use was accessing `pod.Namespace` which is already known to us, that's how we got the pod in the first place.


NetanelBollag:
The call for getting the pod is redundant and made only to verify we received the right pod name and namespace. This can be done better in another location during init, with a proper flag to override for local running on debug.

PR #188, opened 2023-05-23
Read version from a file that can be set from the docker build

omris94:
### Description

- Read the version from a file that can be set from the docker build. 
- Use it for telemetry.
- Send also client_id if defined.


.github/workflows/release-tag.yaml
```diff
@@ -26,18 +26,33 @@ jobs:
...
         uses: docker/login-action@v1
         with:
           username: otterize
           password: ${{ secrets.DOCKER_PASSWORD }}
 
+
       - name: Push to Docker Hub
         run: |-
-          docker pull ${{ env.REGISTRY }}/intents-operator:${{ github.sha }}
+          docker pull $${{ env.REGISTRY }}/intents-operator:${{ github.ref_name }}
```
orishoshan, line 52: Extra $ here

src/operator/Makefile
```diff
@@ -78,7 +78,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	go test ../exp/... -coverprofile cover.out
+	#go test ../exp/... -coverprofile cover.out
```
orishoshan, line 81: !

src/release.Dockerfile
```diff
@@ -0,0 +1,9 @@
+# The only purpose this Dockerfile serves is to be able to run buildx to push a the tag to the version file of the image without rebuilding.
```
orishoshan, line 1: Fix the comment

orishoshan (approved):

PR #200, opened 2023-06-22
Replace dots with underscore in discovered service name

NetanelBollag:
The name of the Otterize service is derived from the K8s resource that owns a pod we want to secure. The name of the resource that owns the pod doesn't necessarily have to be a valid DNS name, but it must follow the [RFC 1123 subdomain naming convention](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names).

This means that '.' is a valid character in the deployment name, which poses a challenge when parsing server names and namespaces in client intents format. The dot is used as a separator between the name and the namespace. For example, if the deployment name is `my-application.4.5.0` in the `infra` namespace, we won't be able to tell if `my-application.4.5.0.infra` is a namespaced name or just the name of the deployment without any namespace.

To address this issue, this PR replaces every dot discovered in the owner resource name with an underscore. In the above example, we would expect the Otterize service name of the mentioned pod to be `my-application_4_5_0.infra`, while `my-application_4_5_0_infra` would be another service name without a namespace.

### DIFF
--- a/src/shared/serviceidresolver/serviceidresolver_test.go
+++ b/src/shared/serviceidresolver/serviceidresolver_test.go
@@ -8,7 +8,10 @@ import (
 	serviceidresolvermocks "github.com/otterize/intents-operator/src/shared/serviceidresolver/mocks"
 	"github.com/stretchr/testify/suite"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


src/shared/serviceidresolver/serviceidresolver_test.go
```diff
@@ -129,6 +132,119 @@ func (s *ServiceIdResolverTestSuite) TestGetPodAnnotatedName_PodMCallFailed() {
...
+			Name:      podName,
+			Namespace: podNamespace,
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					Kind:       "Deployment",
+					Name:       deploymentName,
+					APIVersion: "apps/v1",
+				},
+			},
+		},
```
sapirwo, line 193: you could add some test cases where you expect it to fail (a comment for all the tests)

sapirwo (approved):
you can add some

orishoshan:
Before releasing this, make sure to update the serviceidresolver in the network mapper and the credentials operator so they resolve service names the same way.

orishoshan (approved):

PR #201, opened 2023-06-23
Fixed bug where external traffic policies would be erroneously removed when intents from multiple namespaces were deleted

orishoshan:
1. Fixed bug where external traffic policies would be erroneously removed when intents from multiple namespaces existed and all intents in one namespace were deleted.
2. The experimental flag, `autoCreateNetworkPoliciesForExternalTrafficDisableIntentsRequirement`, was not taken into account during intent deletion flow, so even when it was enabled, removing all intents would result in external traffic network policies being removed.

### DIFF
--- a/src/watcher/reconcilers/pods.go
+++ b/src/watcher/reconcilers/pods.go
@@ -271,7 +271,7 @@ func (p *PodWatcher) InitIntentsServerIndices(mgr ctrl.Manager) error {
 			}
 
 			for _, intent := range intents.GetCallsList() {
-				res = append(res, intent.Name)
+				res = append(res, intent.GetServerFullyQualifiedName(intent.GetServerFullyQualifiedName(intents.Namespace)))
 			}
 
 			return res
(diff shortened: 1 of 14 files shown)


src/watcher/reconcilers/pods.go
```diff
@@ -271,7 +271,7 @@ func (p *PodWatcher) InitIntentsServerIndices(mgr ctrl.Manager) error {
 			}
 
 			for _, intent := range intents.GetCallsList() {
-				res = append(res, intent.Name)
+				res = append(res, intent.GetServerFullyQualifiedName(intent.GetServerFullyQualifiedName(intents.Namespace)))
```
NetanelBollag, line 274: I guess the double call to `GetServerFullyQualifiedName` is a bug. 
Even if not I suggest a better practice of assigning to a local variable. Less cognitive load on the reader and shorter lines:

```suggestion
				fullName := intent.GetServerFullyQualifiedName(intents.Namespace)
				res = append(res, fullName)
```

NetanelBollag (approved):

PR #202, opened 2023-06-28
Detect pods with external network policy

orishavit:
### Description
Detect services for which the intent operator has created external network policy. These services should be marked to allow traffic from unknown sources, as the IPs, or components used by Ingress, for example, are not know in advanced.


src/operator/controllers/external_traffic/network_policy_reconciler.go
```diff
@@ -0,0 +1,150 @@
...
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1.NetworkPolicy{}).
+		WithOptions(controller.Options{RecoverPanic: lo.ToPtr(true)}).
+		WithEventFilter(filterOtterizeNetworkPolicy()).
+		Complete(r)
+}
+
+func (r *NetworkPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logrus.WithField("policy", req.NamespacedName.String()).Infof("Reconcile NetworkPolicy")
```
orishoshan, line 58: Make this debug-level - it'll be spammy

src/operator/main.go
```diff
@@ -193,6 +193,11 @@ func main() {
 		logrus.Infof("Running with enforcement disabled globally, won't perform any enforcement")
 	}
 
+	netpolReconciler := external_traffic.NewNetworkPolicyReconciler(mgr.GetClient(), mgr.GetScheme(), otterizeCloudClient)
+	if err = netpolReconciler.SetupWithManager(mgr); err != nil {
+		logrus.WithError(err).Fatal("unable to init index for netpol")
```
orishoshan, line 198: Fix the log message here

orishoshan (approved):

PR #207, opened 2023-07-17
Ignore policies for non protected

NetanelBollag:
Ignore policies for un-protected services


src/operator/controllers/intents_reconcilers/network_policy.go
```diff
@@ -154,6 +165,28 @@ func (r *NetworkPolicyReconciler) handleNetworkPolicyCreation(
 	return r.UpdateExistingPolicy(ctx, existingPolicy, newPolicy, intent, intentsObjNamespace)
 }
 
+func (r *NetworkPolicyReconciler) shouldProtectServer(ctx context.Context, intent otterizev1alpha2.Intent, intentObjNamespace string) (bool, error) {
```
omris94, line 168: Can you change the signature to get otterizeServiceName instead of intent and ns?

src/operator/controllers/intents_controller.go
```diff
@@ -102,6 +114,46 @@ func (r *IntentsReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				Name:      clientIntents.Name,
+				Namespace: clientIntents.Namespace,
+			},
+		}
+		requests = append(requests, request)
+	}
+	return requests
+}
+
+func (r *IntentsReconciler) getIntentsToProtectedServices(protectedServices *otterizev1alpha2.ProtectedServices) []otterizev1alpha2.ClientIntents {
```
omris94, line 139: What happens if a server that once was a protected service will be removed from the list? Do we catch this scenario?

src/operator/controllers/intents_reconcilers/network_policy.go
```diff
@@ -154,6 +167,31 @@ func (r *NetworkPolicyReconciler) handleNetworkPolicyCreation(
 	return r.UpdateExistingPolicy(ctx, existingPolicy, newPolicy, intent, intentsObjNamespace)
 }
 
+func (r *NetworkPolicyReconciler) shouldProtectServer(ctx context.Context, serverName string, serverNamespace string) (bool, error) {
+	if !viper.GetBool(operatorconfig.EnableProtectedServicesKey) {
+		return true, nil
+	}
+
+	var protectedServicesResources otterizev1alpha2.ProtectedServicesList
+	err := r.List(ctx, &protectedServicesResources, client.InNamespace(serverNamespace))
```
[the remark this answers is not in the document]
NetanelBollag: To index by namespace? What's the benefit?

src/operator/controllers/intents_controller.go
NetanelBollag: Well, I found out we don't delete network policies when client intents are deleted, and this problem should be solved together. We should be able to list all network policies of servers in the namespace and delete the redundants

NetanelBollag: We discussed that offline and this will be solved in another PR

src/operator/controllers/intents_reconcilers/network_policy.go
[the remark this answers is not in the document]
NetanelBollag: So each list will have an index with every existing item on the list? It feels too complicated for me, the for loop is O(n) when n is really, really small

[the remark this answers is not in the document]
NetanelBollag: Let me know if you have more resistance to it and why, but I feel good with the list

omris94 (approved):

PR #211, opened 2023-07-19
Remove orphan network policies

NetanelBollag:
We should delete network policies whenever the intent is removed from the client intents


src/operator/controllers/intents_reconcilers/network_policy.go
```diff
@@ -317,6 +328,66 @@ func (r *NetworkPolicyReconciler) handleNetworkPolicyRemoval(
 	return nil
 }
 
+func (r *NetworkPolicyReconciler) removeOrphanNetworkPolicies(ctx context.Context) error {
+	logrus.Info("Searching for orphaned network policies")
+	networkPolicyList := &v1.NetworkPolicyList{}
+	selector, err := r.matchAccessNetworkPolicy()
```
omris94, line 334: We can index NetworkPolicies by its "from namespace selector". This way we will be able to list only policies that were referenced by intents from the reconciled intent's namespace

src/operator/controllers/intents_reconcilers/network_policy.go
```diff
@@ -117,6 +118,12 @@ func (r *NetworkPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	err = r.removeOrphanNetworkPolicies(ctx)
+	if err != nil {
+		r.RecordWarningEventf(intents, ReasonRemovingNetworkPolicyFailed, "could not remove orphan network policies: %s", err.Error())
```
orishoshan, line 123: What do these network policies represent from the user's perspective? Are they network policies created by intents? Please change the event text to a message that the user could make sense of without context of the implementation (orphan network policies is an implementation detail)

orishoshan, line 123:
```suggestion
		r.RecordWarningEventf(intents, ReasonRemovingNetworkPolicyFailed, "failed to remove network policies: %s", err.Error())
```
Suggest omitting 'orphan' since it's not meaningful to the user (it's an implementation detail). This will remove network policies that are no longer referred to by the current set of intents

orishoshan (approved):

PR #215, opened 2023-07-27
Create protected services reconciler group

NetanelBollag:
To enable adding new reconcilers for protected services, this PR moves the default deny logic to a new reconciler and create the group.

### DIFF
--- a/src/operator/controllers/protected_services_reconcilers/default_deny.go
+++ b/src/operator/controllers/protected_services_reconcilers/default_deny.go
@@ -1,0 +1,171 @@
+package protected_services_reconcilers
+
+import (
+	"context"
+	"fmt"
+	otterizev1alpha2 "github.com/otterize/intents-operator/src/operator/api/v1alpha2"
+	"github.com/otterize/intents-operator/src/shared/injectablerecorder"
+	"github.com/otterize/intents-operator/src/shared/operatorconfig"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+	v1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"reflect"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// DefaultDenyReconciler reconciles a ProtectedServices object
+type DefaultDenyReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+	injectablerecorder.InjectableRecorder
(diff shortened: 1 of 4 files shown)


src/operator/controllers/protected_services_reconcilers/default_deny.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *DefaultDenyReconciler) blockAccessToServices(ctx context.Context, ProtectedServicesResources otterizev1alpha2.ProtectedServicesList, namespace string) error {
+	serversToProtect := map[string]v1.NetworkPolicy{}
+	for _, list := range ProtectedServicesResources.Items {
+		if list.DeletionTimestamp != nil {
+			logrus.Infof("Protected service %s is being deleted and ignored", list.Name)
```
orishoshan, line 57: This doesn't actually ignore the protected service - it just doesn't add it to the list of servers to protect for the calculation later. This message confused me since I was wondering how it then gets deleted.

But I think you can remove the log altogether - this is part of normal functioning, shouldn't have to log this. What do you think?

src/operator/controllers/protected_services_reconcilers/default_deny.go
```diff
@@ -0,0 +1,171 @@
...
+		otterizev1alpha2.OtterizeNetworkPolicyServiceDefaultDeny: "true",
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, existingPolicy := range networkPolicies.Items {
+		existingPolicyServerName := existingPolicy.Labels[otterizev1alpha2.OtterizeNetworkPolicy]
+		_, found := serversToProtect[existingPolicyServerName]
+		if found {
```
orishoshan, line 79: Is it possible to rewrite this for loop to have less nested sections? For example, by checking a condition, then `continue`, instead of doing if..else.

NetanelBollag: This is all an already merged code, I just moved it around

src/operator/controllers/protected_services_reconcilers/default_deny.go
NetanelBollag: This is all an already merged code, I just moved it around

orishoshan (approved):

src/operator/controllers/protected_services_reconcilers/default_deny.go
NetanelBollag: But anyway, `continue` is a problematic feature for those cases it tends to blend the code together and it's tough to maintain it, we have it in the mapper that kind of chaos for loops. Moving it to a separate function will make it even harder to read with same line of code

PR #226, opened 2023-08-08
Create ProtectedService CRD if missing, add webhook to validate duplicate protected services

orishoshan:
This PR makes the operator create the ProtectedService CRD if it is missing. This is a temporary measure, as we switch from using Helm charts to manage our CRDs to asking users to apply CRDs prior to installation. Once we do so and that has been released for a while, we will remove this workaround.

Also in this PR: missing webhook to validate that there are no duplicate protected services

### Testing

### Checklist

### DIFF
--- a/src/operator/protectedservicescrd/applyProtectedServicesCRD.go
+++ b/src/operator/protectedservicescrd/applyProtectedServicesCRD.go
@@ -29,7 +29,7 @@ func EnsureProtectedServicesCRD(ctx context.Context, client client.Client) error
 	if k8serrors.IsNotFound(err) {
 		err := client.Create(ctx, &crdToCreate)
 		if err != nil {
-			return fmt.Errorf("could not get protected services CRD: %w", err)
+			return fmt.Errorf("could not create protected services CRD: %w", err)
 		}
 		return nil
 	}
--- a/src/operator/protectedservicescrd/applyProtectedServicesCRD.go
+++ b/src/operator/protectedservicescrd/applyProtectedServicesCRD.go
@@ -1,0 +1,38 @@
+package protectedservicescrd
+
+import (
+	"context"
+	_ "embed"
+	"fmt"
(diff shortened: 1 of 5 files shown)


src/operator/protectedservicescrd/applyProtectedServicesCRD.go
```diff
@@ -0,0 +1,38 @@
...
+	crd := apiextensionsv1.CustomResourceDefinition{}
+	err = client.Get(ctx, types.NamespacedName{Name: crdToCreate.Name}, &crd)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return fmt.Errorf("could not get protected services CRD: %w", err)
+	}
+
+	if k8serrors.IsNotFound(err) {
+		err := client.Create(ctx, &crdToCreate)
+		if err != nil {
+			return fmt.Errorf("could not get protected services CRD: %w", err)
```
evyatarmeged, line 32: Change logged error

orishoshan: Nice catch

src/operator/webhooks/protectedservices_webhook.go
```diff
@@ -76,6 +76,15 @@ func (v *ProtectedServiceValidator) ValidateUpdate(ctx context.Context, oldObj,
 	var allErrs field.ErrorList
 	protectedServices := newObj.(*otterizev1alpha2.ProtectedService)
 
+	protectedServicesList := &otterizev1alpha2.ProtectedServiceList{}
+	if err := v.List(ctx, protectedServicesList, &client.ListOptions{Namespace: protectedServices.Namespace}); err != nil {
+		return err
+	}
+
+	if err := v.validateNoDuplicateClients(protectedServices, protectedServicesList); err != nil {
```
evyatarmeged, line 84: Need to call this on `ValidateCreate` as well

orishoshan: whoops

evyatarmeged (approved):

PR #235, opened 2023-08-17
Add ProtectedService support for Kafka and Istio, and merge pod watcher into intents operator.

orishoshan:
### Description

1. In the previous release, we added support for ProtectedService resources for network policies. This PR adds initial support for ProtectedService for Kafka and Istio. The initial support only makes ProtectedService enable enforcement, unlike network policies where it also adds a default deny.
2. The intents operator had two components: the operator itself and the pod watcher. Originally, this was done so that the memory consumption of the pod watcher could be separated from the rest of the operator logic. However, this is no longer possible due to changes in the operator, so to simplify things, we are merging them into one component

### Checklist


src/operator/controllers/intents_reconcilers/protected_services/should_protect.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"context"
+	otterizev1alpha2 "github.com/otterize/intents-operator/src/operator/api/v1alpha2"
+	"github.com/sirupsen/logrus"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func ShouldCreateNetworkPoliciesDueToProtectionOrDefaultState(ctx context.Context, kube client.Client, serverName string, serverNamespace string, enforcementDefaultState bool) (bool, error) {
```
evyatarmeged, line 12:
```suggestion
func ShouldCreateNetworkPolicies(ctx context.Context, kube client.Client, serverName string, serverNamespace string, enforcementDefaultState bool) (bool, error) {
```

src/operator/controllers/istiopolicy/policy_manager.go
```diff
@@ -316,14 +318,32 @@ func shouldUpdateStatus(intents v1alpha2.ClientIntents, currentName string, curr
...
+			return nil, err
+		}
+
+		if !shouldCreatePolicy {
+			logrus.Infof("Enforcement is disabled globally and server is not explicitly protected, skipping network policy creation for server %s in namespace %s", intent.GetServerName(), intent.GetServerNamespace(clientIntents.Namespace))
+			c.recorder.RecordNormalEventf(clientIntents, consts.ReasonEnforcementDefaultOff, "Enforcement is disabled globally and called service '%s' is not explicitly protected using a ProtectedService resource, network policy creation skipped", intent.Name)
+			continue
+		}
+
+		if !c.enableIstioPolicyCreation {
```
evyatarmeged, line 342: Why is this a separate flag ?

src/operator/main.go
```diff
@@ -290,6 +295,29 @@ func main() {
 		logrus.WithError(err).Fatal("unable to create controller", "controller", "ProtectedServices")
 	}
 
+	podWatcher := pod_reconcilers2.NewPodWatcher(mgr.GetClient(), mgr.GetEventRecorderFor("intents-operator"), watchedNamespaces, enforcementConfig.EnforcementDefaultState, enforcementConfig.EnableIstioPolicy)
```
evyatarmeged, line 298: rename `pod_reconcilers2`

evyatarmeged (approved):

src/operator/controllers/istiopolicy/policy_manager.go
orishoshan: This is the flag that disabled Istio policy creation, like the one we have for netpols and Kafka

PR #240, opened 2023-08-22
Support user specified annotation for service name override

omris94:
### Description

Following https://github.com/otterize/network-mapper/issues/128 , adding support for user-specified annotation for service name override. The default will be `intents.otterize.com/service-name` as it used to be, and it will be configurable through the environment variable: `OTTERIZE_SERVICE_NAME_OVERRIDE_ANNOTATION`. 

After this PR we should update `go.mod` for credentials-operator and network-mapper, And then expose the new env var as an helm chart value. 

### References

https://github.com/otterize/network-mapper/issues/128

### Testing

### Checklist

### DIFF
--- a/src/shared/serviceidresolver/config.go
+++ b/src/shared/serviceidresolver/config.go
@@ -1,0 +1,19 @@
+package serviceidresolver
+
+import (
+	"github.com/spf13/viper"
+	"strings"
+)
+
+const (
+	serviceNameOverrideAnnotationKey        = "service-name-override-annotation"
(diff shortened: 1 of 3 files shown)


src/shared/serviceidresolver/config.go
```diff
@@ -0,0 +1,19 @@
+package serviceidresolver
+
+import (
+	"github.com/spf13/viper"
+	"strings"
+)
+
+const (
+	serviceNameOverrideAnnotationKey        = "service-name-override-annotation"
```
orishoshan, line 9: Config goes through commandline args in the operators - see main.go

PR #244, opened 2023-09-07
Send applied intents telemetry

NetanelBollag:
### Testing


src/operator/controllers/intents_controller.go
```diff
@@ -115,6 +122,36 @@ func (r *IntentsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	return r.group.Reconcile(ctx, req)
 }
 
+func (r *IntentsReconciler) intentsReconcilerInit(ctx context.Context) error {
+	if !telemetrysender.IsTelemetryEnabled() {
```
NetanelBollag, line 126: @orishoshan Have a look here. The thumb rule is that every reconciler that has a finalizer should always run since we always must remove the finalizer even when no other logic is required. We must have a finalizer to decrement the counter on deletion of resource. In this case I preferred to just delete all finalizers on init if telemetries disabled, so it will be very clear that this reconciler doesn't run at all.

evyatarmeged (approved):

PR #246, opened 2023-09-09
Fix network policy cleanup from unprotected services

omris94:
### Description

Fix network policy cleanup from unprotected services by considering protectedServices in a deletion state as unprotected. 

The method `CleanPoliciesFromUnprotectedServices` was used to make a list of *all* the services that a ProtectedService resource refers to. The problem with that approach is that a resource might be in a deletion state. If so, we would expect that its service will be considered unprotected.

### Testing

Added a test for this scenario.


src/operator/controllers/intents_reconcilers/network_policy.go
```diff
@@ -383,8 +383,11 @@ func (r *NetworkPolicyReconciler) CleanPoliciesFromUnprotectedServices(ctx conte
 
 	protectedServersByNamespace := sets.Set[string]{}
 	for _, protectedService := range protectedServicesResources.Items {
-		serverName := otterizev1alpha2.GetFormattedOtterizeIdentity(protectedService.Spec.Name, namespace)
-		protectedServersByNamespace.Insert(serverName)
+		// skip protected services that are in deletion process
+		if protectedService.DeletionTimestamp == nil {
```
orishoshan, line 387: This checks that they are not being deleted.

Instead, make it like an err check, negate the check. If they are being deleted (!= nil), then `continue`

orishoshan (approved):

PR #248, opened 2023-09-16
Fixed ProtectedService bugs: global network policy enforcement setting, and KafkaServerConfig cleanup

orishoshan:
Fixed bug where ProtectedService would ignore setting for network policy enforcement
Fixed bug where ProtectedService would not cause KafkaServerConfigs to remove their ACLs

### DIFF
--- a/src/operator/controllers/protected_service_reconcilers/default_deny_test.go
+++ b/src/operator/controllers/protected_service_reconcilers/default_deny_test.go
@@ -44,13 +44,56 @@ func (s *DefaultDenyReconcilerTestSuite) SetupTest() {
 	s.MocksSuiteBase.SetupTest()
 
 	s.extNetpolHandler = protectedservicesmock.NewMockExternalNepolHandler(s.Controller)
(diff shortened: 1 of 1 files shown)


src/operator/controllers/kafkaserverconfig_controller.go
```diff
@@ -332,7 +335,7 @@ func (r *KafkaServerConfigReconciler) uploadKafkaServerConfigs(ctx context.Conte
 }
 
 func kafkaServerConfigCRDToCloudModel(kafkaServerConfig otterizev1alpha2.KafkaServerConfig) (graphqlclient.KafkaServerConfigInput, error) {
-	var topics []graphqlclient.KafkaTopicInput
+	topics := make([]graphqlclient.KafkaTopicInput, 0)
```
orishoshan, line 338: Fixes bug where if no `topics` were specified in the spec, the cloud reconciler would fail to report this because it was `nil` (empty list is ok, nil is not)

omris94 (approved):

PR #255, opened 2023-09-28
Create unique event counter for telemetries

NetanelBollag:
Unique event counter allow to count distinct events and send them as telemetry while using only anonymized and hashed data and send only the distinct count of the event. 

### Checklist

### DIFF
--- a/src/shared/telemetries/telemetrysender/config.go
+++ b/src/shared/telemetries/telemetrysender/config.go
@@ -6,17 +6,19 @@ import (
 )
 
 const (
-	TelemetryAPIAddressKey       = "telemetry-address"
-	TimeoutKey                   = "telemetry-client-timeout"
-	CloudClientTimeoutDefault    = "30s"
-	TelemetryEnabledKey          = "telemetry-enabled"
-	TelemetryEnabledDefault      = false
-	TelemetryMaxBatchSizeKey     = "telemetry-max-batch-size"
-	TelemetryMaxBatchSizeDefault = 100
-	TelemetryIntervalKey         = "telemetry-interval-seconds"
-	TelemetryIntervalDefault     = 5
-	TelemetryAddressDefault      = "https://app.otterize.com/api/telemetry/query"
-	EnvPrefix                    = "OTTERIZE"
+	TelemetryAPIAddressKey                = "telemetry-address"
+	TimeoutKey                            = "telemetry-client-timeout"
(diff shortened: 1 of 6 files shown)


src/shared/telemetries/telemetrysender/sender.go
```diff
@@ -55,6 +70,58 @@ func (t *TelemetrySender) Send(component telemetriesgql.Component, eventType tel
...
+
+	for _, item := range batch {
+		t.uniqueEventsCounter.IncrementCounter(item.Event.Component, item.Event.EventType, item.Key)
+	}
+
+	counts := t.uniqueEventsCounter.Get()
+	logrus.Infof("Sending telemetry snapshot: %v", counts)
+	timeUntilReset := t.lastSnapshotResetTime.Add(t.snapshotResetInterval)
+
+	if time.Now().After(timeUntilReset) {
```
omris94, line 109: I think it should be at the end of the function 🤔

src/shared/telemetries/telemetrysender/config.go
```diff
@@ -6,17 +6,19 @@ import (
...
+	TimeoutKey                            = "telemetry-client-timeout"
+	CloudClientTimeoutDefault             = "30s"
+	TelemetryEnabledKey                   = "telemetry-enabled"
+	TelemetryEnabledDefault               = false
+	TelemetryMaxBatchSizeKey              = "telemetry-max-batch-size"
+	TelemetryMaxBatchSizeDefault          = 100
+	TelemetryIntervalKey                  = "telemetry-interval-seconds"
+	TelemetryIntervalDefault              = 5
+	TelemetryAddressDefault               = "https://app.otterize.com/api/telemetry/query"
+	TelemetrySnapshotResetIntervalKey     = "telemetry-snapshot-reset-interval-duration"
```
omris94, line 19: I think the name "Snapshot telemetry" might be a little bit intimidating.

src/shared/telemetries/telemetrysender/sender.go
NetanelBollag: 100%, also I abort the reset if the send fail

src/shared/telemetries/telemetrysender/config.go
NetanelBollag: Fixed it to reset which is actually what it is

omris94 (approved):

PR #256, opened 2023-10-01
Adding support for using Kubernetes services as target server in intents, adding port restriction in the process.

evyatarmeged:
### Description
This PR adds experimental support for  Kubernetes services in the intents `calls:` section (target servers). You can specify a Kubernetes service in this fashion:
```yaml
calls:
  - name: svc:k8s-service
```
This causes the operator to review the service's target ports, restricting the network policy to these ports. When specified without this svc qualifier, then the network policy opens all ports.

This is experimental and the ClientIntents format will change to accommodate this new style of addressing services.

Accompanying pull request in the network-mapper https://github.com/otterize/network-mapper/pull/145 allows auto-detecting traffic to Kubernetes services and generating ClientIntents for them.


src/operator/controllers/intents_reconcilers/network_policy.go
```diff
@@ -463,13 +476,56 @@ func (r *NetworkPolicyReconciler) buildNetworkPolicyObjectForIntent(
...
+	}, &svc)
+
+	if err != nil {
+		return err
+	}
+
+	portToProtocol := make(map[int]corev1.Protocol)
+	// Gather all target ports (target ports in the pod the service proxies to)
+	for _, port := range svc.Spec.Ports {
+		portToProtocol[port.TargetPort.IntValue()] = port.Protocol
```
orishoshan, line 515: What if TargetPort is specified as string? We can note that it's unsupported here (you can provide the port's "name", and not the number, which would mean it's specified in the Pod)

evyatarmeged: We can fetch the pod with the Service's label selector and convert the string to int. Tell me WDYT

orishoshan: But then we would have to reconcile pods too, no? I think better to leave this as a limitation for now, and if there is no target port as int, send a Kubernetes Event Warning that no port was used because named ports are unsupported.

src/operator/controllers/svc_reconcilers/svc_reconcilers.go
```diff
@@ -0,0 +1,105 @@
...
+func (p *ServiceWatcher) Register(mgr manager.Manager) error {
+	watcher, err := controller.New("otterize-service-watcher", mgr, controller.Options{
+		Reconciler:   p,
+		RecoverPanic: lo.ToPtr(true),
+	})
+	if err != nil {
+		return fmt.Errorf("unable to set up service controller: %p", err)
+	}
+
+	if err = watcher.Watch(&source.Kind{Type: &corev1.Service{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
orishoshan, line 74: This is equivalent to:
```	return ctrl.NewControllerManagedBy(mgr).
		For(&corev1.Service{}).
		WithOptions(controller.Options{RecoverPanic: lo.ToPtr(true)}).
		Complete(r)
```

Which is the recommended syntax, calling Watch directly is the more low-level syntax that you should use if you are trying to do something a little bit different (like we do elsewhere with Watch)

src/operator/controllers/svc_reconcilers/svc_reconcilers.go
```diff
@@ -0,0 +1,153 @@
...
+	err := r.Get(ctx, req.NamespacedName, &service)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return ctrl.Result{}, err
+	}
+
+	targetServerName := fmt.Sprintf("svc:%s.%s", req.Name, req.Namespace)
+	var intentsList otterizev1alpha2.ClientIntentsList
+	err = r.List(
+		ctx, &intentsList,
+		&client.MatchingFields{otterizev1alpha2.OtterizeTargetServerIndexField: targetServerName})
```
orishoshan, line 50: What about svc:svcname without namespace?

orishoshan: Should List for ClientIntents in same namespace of Service that refer to service name without the namespace suffix.

orishoshan (approved):

PR #261, opened 2023-10-09
Replaced multiple finalizers placed on Otterize CRDs with a single finalizer per resource type

NetanelBollag:
### Description

Remove the duplicated finalizers on Otterize CRDs and handle all the finalizers logic inside the reconcilers group. 
Particular reconcilers aren't required to be aware of their finalizers logic anymore. 

### How the reconciler group will handle finalizers
Finalizers will be add to resources before any reconciler logic will run.

If a reconciler have a specific logic for resource deletion, like removing policies, they need to check if the resource is being deleted and return after they are done. If all reconcilers returned without error, the resource's finalizer will be removed.

### Legacy finalizers 
For backward compatibility with existing operator, the reconcilers group will delete every legacy finalizers from resources on init.


src/operator/controllers/intents_controller.go
```diff
@@ -133,26 +148,25 @@ func (r *IntentsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 
 func (r *IntentsReconciler) intentsReconcilerInit(ctx context.Context) error {
-	if !telemetrysender.IsTelemetryEnabled() {
-		// When telemetry is disabled no logic should run by the telemetry reconciler. In case that a previous run of
-		// the operator had telemetry enabled we must remove the telemetry reconciler finalizers from all the CRDs.
-		err := r.RemoveFinalizerFromAllResources(ctx, otterizev1alpha2.OtterizeTelemetryReconcilerFinalizerName)
-		if err != nil {
-			return err
-		}
+	err := r.RemoveLegacyFinalizer(ctx)
```
orishoshan, line 151: RemoveLegacyFinalizer happens here and removes all finalizers for all resources, but also in the reconcilergroup per-resource (which is the correct behavior)

orishoshan (approved):

PR #262, opened 2023-10-10
Use namespace and serviceAccount name when creating AWS roles

orishavit:
Prior to this commit, the existing code used only the serviceAccount's name, instead of a name and namespace. This change will include the namespace in the created AWS role name, and also tags the role with the namespace name, to be checked before deleting the role.


src/shared/awsagent/roles.go
```diff
@@ -98,23 +103,42 @@ func (a *Agent) DeleteOtterizeIAMRole(ctx context.Context, accountName string) e
...
-	_, err = a.iamClient.DeleteRolePolicy(ctx, &iam.DeleteRolePolicyInput{
+	_, err = a.iamClient.GetRolePolicy(ctx, &iam.GetRolePolicyInput{
 		PolicyName: role.RoleName,
 		RoleName:   role.RoleName,
 	})
 
-	if err != nil {
-		logger.WithError(err).Errorf("failed to delete inline policy")
-		return err
+	if err == nil {
```
orishoshan, line 127: Reverse the condition - early return using err != nil instead of nested conditions

src/shared/awsagent/roles.go
```diff
@@ -98,23 +103,42 @@ func (a *Agent) DeleteOtterizeIAMRole(ctx context.Context, accountName string) e
...
+		_, err = a.iamClient.DeleteRolePolicy(ctx, &iam.DeleteRolePolicyInput{
+			PolicyName: role.RoleName,
+			RoleName:   role.RoleName,
+		})
+
+		if err != nil {
+			logger.WithError(err).Errorf("failed to delete inline policy")
+			return err
```
orishoshan, line 135: If an error is returned, then don't log - let the logger above you handle logging (in this case it's controller-runtime's logger). If you want to add a custom message, use `fmt.Errorf("my custom message for error: %w", err)`, which will add a custom string to the error and also wrap the message so `errors.Is` and `errors.As` still works.

omris94 (approved):

Issue #249, opened 2023-09-25
Add Status field to ClientIntents


orishoshan:
Many resources in Kubernetes have a Status field that indicates the state of the resource.

ClientIntents are reconciled and as this is done, different kinds of policies may be created (or fail to create) depending on a variety of reasons. Today, these reasons are reflected only in Kubernetes Events.

Proposal:
Add a Kubernetes Status field to ClientIntents, which has a 'UP TO DATE' field, as well as a 'ENFORCED' field for every kind of policy (e.g. network policies, Kafka ACLs, Istio policies).

When a ClientIntent begins reconciliation, its UP TO DATE field should be set to false. When reconciliation completes successfully, UP TO DATE should be set to true.

The ENFORCED fields for each kind of policy should be set to true only if the policy is being enforced, regardless of whether it's due to an error, the feature is disabled, or the intents operator is in shadow mode (meaning enforcement is disabled by default).

This should not replace the Kubernetes Events that are sent.

aerosouund:
Hey, i wanna take on this. I will attempt to check out the code and ask here if i need more context on something

aerosouund:
I've took a tour in the code and saw that regarding the status field the intents type definition already has an `IntentsStatus` field. an UpToDate field can be added and this field will be accessed by the `Reconcile` method implementation for each reconciler. Am i correct in this line of thinking ? let me know if there's any other part that needs to be done

and i will check as well what needs to be done for the enforced policy proposal

aerosouund:
I've took a tour in the code and saw that regarding the status field the intents type definition already has an `IntentsStatus` field. an UpToDate field can be added and this field will be accessed by the `Reconcile` method implementation for each reconciler. Am i correct in this line of thinking ? let me know if there's any other part that needs to be done

and i will check as well what needs to be done for the enforced policy proposal

orishoshan:
Hey @aerosouund! Yep, you're spot on. Feel free to reach out to me on Slack if you need assistance :)
Sorry for the late reply - notifications weren't set up right for this repo. Fixed it now so we get notified for issue comments as well :)

aerosouund:
Hello @orishoshan 
Please review the PR and let me know. i also didn't know how to test this so please let me know if there's a way you'd recommend.
If you also prefer to leave the PR open until i add the enforced field

orishoshan:
Completed in #266

PR #266, opened 2023-10-16
add uptodate field and reconcile logic

aerosouund:
### Description

Add UpToDate field to IntentsStatus struct. update this field once intent processing has completed

### References

ISSUE: https://github.com/otterize/intents-operator/issues/249


orishoshan:
Hey @aerosouund , thank you!

src/operator/controllers/intents_reconcilers/kafka_acls.go
```diff
@@ -190,6 +191,13 @@ func (r *KafkaACLReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	}
 
 	var result ctrl.Result
+
+	intents.Status.UpToDate = false
+
+	defer func() {
+		intents.Status.UpToDate = true
```
orishoshan, line 198: You should set UpToDate = true only if there is no error, so move it above the `return ctrl.Result{}, nil` line.
If you do it it the `defer func()` block, then it will happen anyway, upon return. If an error or panic occurred, then the resource is not up to date.

Also, we need to remember to update the intents resource for this to take effect.

src/operator/controllers/intents_reconcilers/network_policy.go
```diff
@@ -76,6 +77,12 @@ func (r *NetworkPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	logrus.Infof("Reconciling network policies for service %s in namespace %s",
 		intents.Spec.Service.Name, req.Namespace)
 
+	intents.Status.UpToDate = false
+
+	defer func() {
+		intents.Status.UpToDate = true
```
orishoshan, line 83: Let's move this to `intents_controller.go`, which has a `Reconcile` func that calls all the sub-reconcilers. Once they have all succeeded, you can set `UpToDate = true` there, and won't have to duplicate this logic across multiple reconcilers.

orishoshan (changes requested):
@aerosouund Almost there! See minor comments.

src/operator/controllers/intents_reconcilers/network_policy.go
aerosouund: he method there has no awareness of the underlying fields in the structs. i wrote the code in all files because the only place where i can access the struct fields is after i call `err := r.client.Get(ctx, req.NamespacedName, intents)` . or do you think we should make this call in the Reconcile method in the intents_controller file as well ?

src/operator/controllers/intents_reconcilers/kafka_acls.go
aerosouund: Agreed, i will move the call after we have settled on what to do in the other conversation

src/operator/controllers/intents_reconcilers/network_policy.go
amitlicht: Hi there @aerosouund! Jumping in per @orishoshan's request. 
I agree with Ori here - we should avoid duplicating this logic across the different group reconcilers, and instead apply the status update in a single place. This also prevents a possible bug in your current implementation, where the value of UpToDate is ultimately determined only according to the last reconciler execution. 
Indeed, to do so you'd have to call `r.client.Get(...)` at the root reconciler, that is in `IntentsReconciler.Reconcile` function, at intents_reconciller.go. 

If you have any follow-up questions feel free to contact me using comments here, or at our Otterize Community slack.

amitlicht (changes requested):
See my comment in the thread.

src/operator/controllers/intents_reconcilers/network_policy.go
aerosouund: Updated the logic, thanks for the concise explanation

src/operator/controllers/intents_controller.go
```diff
@@ -125,7 +127,23 @@ func (r *IntentsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		}
+		return ctrl.Result{}, err
+	}
+
+	result, err := r.group.Reconcile(ctx, req)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	intents.Status.UpToDate = true
```
amitlicht, line 145: Did you get a chance to test this and see that it's actually reflected in the object status? 
According to the documentation, applying the change only locally on the in-memory object shouldn't be enough. 
`client.Status().Update(...)`, as demonstrated here https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html  and here https://kubebyexample.com/learning-paths/operator-framework/operator-sdk-go/watching-resources

aerosouund: No, one of the things i asked Ori about was how to test this so thanks for pointing me to that. i'll do that and provide an update.

amitlicht: Great. As per how to test it - I think the easiest way would be to set up a k8s cluster (you can do so using minikube) and deploy the operator onto it. LMK if you have any questions about that.

aerosouund: Done and tested, please review the new code

src/operator/controllers/intents_controller.go
```diff
@@ -125,7 +127,30 @@ func (r *IntentsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
 
-	return r.group.Reconcile(ctx, req)
+	intents := &otterizev1alpha2.ClientIntents{
```
amitlicht, line 130: You could just initialize it to an empty ClientIntents struct, i.e.: 
```
intents := &otterizev1alpha2.ClientIntents{}
```
(initializing the status to the default empty value is implied).

amitlicht (approved):
Great! Approved with a small comment. 
Also note the pending merge conflict - do let me know if fixing it leads to any significant changes in your PR that require re-reviewing.

src/operator/controllers/intents_controller.go
aerosouund: Indeed you're right. changed it and resolved the merge conflict

amitlicht:
@aerosouund I'm looking into why the build is failing (this doesn't seem to be related to your changes). Will update.

src/operator/config/crd/k8s.otterize.com_clientintents.yaml
```diff
@@ -132,6 +132,12 @@ spec:
             type: object
           status:
             description: IntentsStatus defines the observed state of ClientIntents
+            properties:
+              upToDate:
+                description: 'INSERT ADDITIONAL STATUS FIELD - define observed state
```
amitlicht, line 137: please fix the description here to describe this new property.

amitlicht (approved):
see comment inside

src/operator/config/crd/k8s.otterize.com_clientintents.yaml
aerosouund: Updated, please check

amitlicht (approved):

aerosouund:
build still failed, is this having to do with my code ?

amitlicht:
@aerosouund congrats and thank you for your contribution! I merged it in. If we have any follow-ups I will let you know.

PR #268, opened 2023-10-18
Fix bug: Uploading empty network policy list when no pod found for the policy

NetanelBollag:
### Description

The Graphql API for reporting external network policies expect non empty network policy list, but in case there is a network policy that its pod is down, we will report the list anyway.

This PR skip the reporting if no pod found for the network policy (which may happen if the pod is down or doesn't exists) or if we failed to resolve an Otterize service for it (less likely but it may happen, only write a warning log)

### Testing


src/operator/controllers/external_traffic/network_policy_uploader.go
```diff
@@ -80,27 +77,32 @@ func (r *NetworkPolicyUploaderReconciler) Reconcile(ctx context.Context, req ctr
...
+			WithField("policy", req.NamespacedName.String()).
+			Debug("Failed to resolve any pods, skipping reporting")
+		return ctrl.Result{}, nil
+	}
+
 	var inputs []graphqlclient.NetworkPolicyInput
 
 	for _, pod := range podList.Items {
 		serviceId, err := r.serviceIdResolver.ResolvePodToServiceIdentity(ctx, &pod)
+		if err != nil {
```
orishoshan, line 96: We should keep retrying here - an error here is never "expected" (unlike 0 results), so it's reasonable to retry. We do want to make sure we don't miss data. Any flow that is critical has to return an error and be retried, unless we expect this to fail for some reason. If a reconciler gives up due to an error - it is missing data.

src/operator/controllers/external_traffic/network_policy_uploader.go
```diff
@@ -109,6 +111,13 @@ func (r *NetworkPolicyUploaderReconciler) Reconcile(ctx context.Context, req ctr
 		})
 	}
 
+	if len(inputs) == 0 {
```
orishoshan, line 114: Is 0 results an unexpected flow that occurred due to an error? If so, it should be an error.
If it is an expected flow (i.e. we might not see pods), then we should log it in Debug level.

src/operator/controllers/external_traffic/network_policy_uploader.go
```diff
@@ -80,27 +77,32 @@ func (r *NetworkPolicyUploaderReconciler) Reconcile(ctx context.Context, req ctr
 		ctx, &podList,
 		&client.MatchingLabelsSelector{Selector: selector},
 		&client.ListOptions{Namespace: netpol.Namespace})
-
 	if err != nil {
 		logrus.WithError(err).Errorf("error when reading podlist")
 		return ctrl.Result{}, nil
 	}
 
+	if len(podList.Items) == 0 {
```
orishoshan, line 85: I think this suggests that we need to also run this logic on the PodReconciler, as network policies may be present without any pods, and the pods may appear later.

NetanelBollag: Agreed, from a short search it seem there are other location in the code who need it, so I suggest we make it a separate task

src/operator/controllers/external_traffic/network_policy_uploader.go
NetanelBollag: agreed, turned this into an error

src/operator/controllers/external_traffic/network_policy_uploader.go
NetanelBollag: since now the error is not skipped, I removed this section.

src/operator/controllers/external_traffic/network_policy_uploader.go
```diff
@@ -80,27 +82,31 @@ func (r *NetworkPolicyUploaderReconciler) Reconcile(ctx context.Context, req ctr
...
+		logrus.
+			WithField("policy", req.NamespacedName.String()).
+			Debug("Failed to resolve any pods, skipping reporting")
```
orishoshan, line 93:
```suggestion
			Debug("Failed to resolve any pods, will retry")
```

src/operator/controllers/external_traffic/network_policy_uploader.go
```diff
@@ -20,6 +20,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"time"
+)
+
+const (
+	retryGettingPodsForPolicy = time.Duration(5) * time.Second
```
orishoshan, line 27:
```suggestion
	listPodsForPolicyRetryDelay = 5 * time.Second
```
Rephrased it, but the main change here that I was trying to get is adding the word "delay"
And also prefer `5 * time.Second` instead of wrapping it with time.Duration - the idiomatic way to write durations in Go is to multiply numbers by the constants in `time` so that it reads like an English sentence, when read from left to right. e.g. 5 * time.Second = 5 seconds.

src/operator/controllers/external_traffic/network_policy_uploader.go
```diff
@@ -80,27 +82,31 @@ func (r *NetworkPolicyUploaderReconciler) Reconcile(ctx context.Context, req ctr
...
 	if err != nil {
 		logrus.WithError(err).Errorf("error when reading podlist")
 		return ctrl.Result{}, nil
 	}
 
+	if len(podList.Items) == 0 {
+		logrus.
+			WithField("policy", req.NamespacedName.String()).
+			Debug("Failed to resolve any pods, skipping reporting")
+		return ctrl.Result{RequeueAfter: retryGettingPodsForPolicy}, nil
```
orishoshan, line 94:
```suggestion
		return ctrl.Result{RequeueAfter: listPodsForPolicyRetryDelay}, nil
```

orishoshan (approved):

src/operator/controllers/external_traffic/network_policy_uploader.go
NetanelBollag: accepted

PR #279, opened 2023-11-05
Add AWS pod watcher, various other small fixes

orishoshan:
### Description

This PR adds an AWS pod watcher, which reconciles ClientIntents for a pod when it starts.

Tests still in progress.

### Testing

### Checklist


src/operator/api/v1alpha3/clientintents_types.go
```diff
@@ -222,9 +222,12 @@ func (in *ClientIntents) GetFilteredCallsList(intentTypes ...IntentType) []Inten
 }
 
 func (in *ClientIntents) GetIntentsLabelMapping(requestNamespace string) map[string]string {
-	otterizeAccessLabels := map[string]string{}
+	otterizeAccessLabels := make(map[string]string)
 
-	for _, intent := range in.GetFilteredCallsList(IntentTypeHTTP, IntentTypeKafka) {
+	for _, intent := range in.GetCallsList() {
```
orishavit, line 227: why not use GetFilteredCallsList?

orishoshan: Because that filters in, and we need a filter out - we also wanna include intents with no type, or this breaks network policies (which require no type)

orishavit (approved):

PR #287, opened 2023-11-09
Fix bug: Empty intents list reported when there are no database intents

NetanelBollag:
### Description

Until this PR, database reconciler reports nil pointer when client intents doesn't contain any database type. This cause backend error since the GraphQL mutation doesn't support nil variable. In the case there are no `Database` typed intents there is no database action to perform, so nothing should be done and we can skip the API call.

### Testing


NetanelBollag:
Not a bug, the schema allow empty lists

amitlicht (approved):

PR #290, opened 2023-11-17
Retrieve OIDC URL & EKS cluster from AWS without configuration

orishavit:
Currently the operators require to manually learn about the cluster's OIDC provider. This commit retrieves the URL from AWS sources. The operator AWS roles will now need the `ec2:DescribeInstances` and `eks:DescribeCluster` permissions. The current configuration parameter will keep working, in case that this information is not available.

I get the OIDC URL by: Getting the current instance identification document by using EC2 Metadata, get the instance ID, `DescribeInstance` to get the instance tags, which have the cluster name, and then `DescribeCluster` to get the current cluster OIDC URL.


src/shared/awsagent/agent.go
```diff
@@ -18,20 +26,31 @@ type Agent struct {
...
 		panic(err)
 	}
 
+	oidcUrl := ""
+
+	if oidcUrlFromEnv == "" {
+		retrieved, err := tryRetrieveOIDCUrl(ctx, awsConfig)
+
+		if err != nil {
+			logrus.WithError(err).Fatal("OIDC URL not provided, and not retrieved")
```
orishoshan, line 45: @orishavit - let's stick to returning errors and only exiting in main on failure. should not use fatal here

src/shared/awsagent/agent.go
```diff
@@ -18,20 +26,31 @@ type Agent struct {
 
 func NewAWSAgent(
 	ctx context.Context,
-	oidcUrl string, // TODO: use eks.DescribeCluster to get the OIDC URL.
+	oidcUrlFromEnv string,
```
orishoshan, line 29: Initials (like OIDC, URL) should use the correct case: https://google.github.io/styleguide/go/decisions#initialisms

orishoshan (approved):

PR #301, opened 2023-11-29
Support event recording for database-typed intents

omris94:
### Description

Record success and failure events when handlind database intentns. Graphql errors with the right type will be added to the event description as is. 

### Testing

### Checklist


src/operator/controllers/intents_reconcilers/exp/database.go
```diff
@@ -68,8 +74,16 @@ func (r *DatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	}
 
 	if err := r.otterizeClient.ApplyDatabaseIntent(ctx, intentInputList, action); err != nil {
+		errType, errMsg, ok := graphqlclient.GetGraphQLUserError(err)
+		if !ok || errType != graphqlclient.UserErrorTypeAppliedIntentsError {
+			r.RecordWarningEventf(intents, ReasonApplyingDatabaseIntentsFailed, "Failed applying database intents")
```
orishoshan, line 79: error should be included in the message, like other warnings, e.g. "Failed applying database intents: %s", err.Error()

src/operator/controllers/intents_reconcilers/exp/database.go
```diff
@@ -68,8 +74,16 @@ func (r *DatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+			return ctrl.Result{}, err
+		}
+		r.RecordWarningEventf(intents, ReasonApplyingDatabaseIntentsFailed, fmt.Sprintf("%s", errMsg))
 		return ctrl.Result{}, err
 	}
 
+	r.RecordNormalEventf(intents, ReasonAppliedDatabaseIntents, "Successfully applied database intents")
```
orishoshan, line 86: How many intents? This is different from other reconcilers, they show a message with the number of intents successfully reconciled.

src/shared/otterizecloud/graphqlclient/errors.go
```diff
@@ -0,0 +1,34 @@
...
+	errList := &(gqlerror.List{})
+	if !errors.As(err, errList) {
+		return "", "", false
+	}
+	gqlErr := &gqlerror.Error{}
+	if !errList.As(&gqlErr) {
+		return "", "", false
+	}
+	errType, ok := gqlErr.Extensions["errorType"].(string)
+	if !ok || !UserErrorType(errType).IsValid() {
```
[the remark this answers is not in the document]
omris94: I don't think it requires an error. If errType is not valid, it's probably because the graphQL is not current. It makes more sense to use `error.Error()` than throwing another kind of error.

PR #310, opened 2023-12-07
Report anonymized data on panic and errors

NetanelBollag:
### Description

Report anonymous data on errors and crashes, only error type and stack trace is sent.

### Testing

This infrastructure change can be tested manually.

### Checklist


src/operator/otterizecrds/clientintents-customresourcedefinition.yaml
```diff
@@ -3,7 +3,6 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.9.0
-  creationTimestamp: null
```
amitlicht, line 6: is this intentional? how is that related to this PR?

src/shared/telemetries/telemetriesconfig/config.go
```diff
@@ -9,8 +9,8 @@ const (
 	TelemetryAPIAddressKey         = "telemetry-address"
 	TimeoutKey                     = "telemetry-client-timeout"
 	CloudClientTimeoutDefault      = "30s"
-	TelemetryEnabledKey            = "telemetry-enabled"
-	TelemetryEnabledDefault        = false
+	TelemetryUsageEnabledKey       = "telemetry-usage-enabled"
```
amitlicht, line 12: I'd keep the old flag as is or add a deprecated alias for it to avoid breaking existing deployments, but please double-check with ori (since it's a product definition question)

amitlicht (approved):
see comments inside

src/shared/telemetries/telemetriesconfig/config.go
NetanelBollag: In the helm I will anyway pass `telemetry-errors-enabled` so it will be broken if the image isn't updated. This will be a totally separated chart version since it's a breaking change, so I'm not worry

NetanelBollag: We will keep the old flag as global enabler for the whole

PR #319, opened 2023-12-20
Add support for specifying intents for Internet IP addresses

NetanelBollag:
### Description

This pull request introduces a new intent type designed to manage access to internet addresses. Specifically, for this intent type, the name field is made optional. Due to the limitations of the CRD schema in enforcing the name field requirement based on intent type, the enforcement of this requirement will be handled in the webhook server logic.

Example for new intent type:

```
apiVersion: k8s.otterize.com/v1alpha3
kind: ClientIntents
metadata:
  name: client2
  namespace: otterize-tutorial-mapper
spec:
  service:
    name: client2
  calls:
    - type: internet
      internet:
        ips:
          - 185.15.59.0/24
          - 10.52.1.12
        ports:
          - 443

```

### References

[Related Helm chart changes](https://github.com/otterize/helm-charts/pull/150)
### Testing

### Checklist

- Documentation: This is part of egress network policy feature which is still experimental and will be documented when the feature is released


src/operator/controllers/intents_reconcilers/internet_network_policy/internet_network_policy.go
```diff
@@ -0,0 +1,368 @@
...
+		r.RecordNormalEventf(intents, consts.ReasonEnforcementDefaultOff, "Enforcement is disabled globally, internet network policy creation skipped")
+		return ctrl.Result{}, nil
+	}
+	if !r.enableNetworkPolicyCreation {
+		logrus.Infof("Network policy creation is disabled, skipping internet network policy creation for service %s in namespace %s", intents.Spec.Service.Name, req.Namespace)
+		r.RecordNormalEvent(intents, consts.ReasonEgressNetworkPolicyCreationDisabled, "Network policy creation is disabled, internet network policy creation skipped")
+		return ctrl.Result{}, nil
+	}
+
+	hasAnyInternetIntents := slices.ContainsFunc(intents.GetCallsList(), func(intent otterizev1alpha3.Intent) bool {
```
orishavit, line 96: I would move this to the top of the function to avoid unnecessary work. You can also use `GetFilteredCallsList`

orishavit (changes requested):

src/operator/controllers/intents_reconcilers/internet_network_policy/internet_network_policy.go
NetanelBollag: Fixed, moved it up but only after handling `removeOrphan` since it should be handled in any case

orishavit (approved):

PR #323, opened 2023-12-28
Introduce a new type: `serviceEffectivePolicy` as a better abstraction than ClientIntents for reconciliation. Implement `EffectivePolicyReconciler` for ingress network policies.

omris94:
### Description

This pull request introduces two new abstractions: `ServiceEffectivePolicy` and `ServerEffectivePolicyGroupReconciler`. These abstractions are designed to help reconcile input resources (such as intents, pods, and protected services) to output policies (such as network policies for Ingress, Kafka ACLs, and Postgres). 

The `ServiceEffectivePolicy` type is created, and an `EffectivePolicyReconciler` is implemented specifically for Ingress network policies.

### Testing

I modified the existing tests of ingress policies (and external netpols) according to the changes.  

### Checklist


src/operator/effectivepolicy/Syncer.go
```diff
@@ -0,0 +1,142 @@
+package effectivepolicy
```
orishoshan, line 1: filename should be lowercase

src/operator/effectivepolicy/Syncer.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/otterize/intents-operator/src/shared/injectablerecorder"
+	"github.com/otterize/intents-operator/src/shared/serviceidresolver/serviceidentity"
+	"github.com/samber/lo"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type EffectivePoliciesApplier interface {
+	ApplyEffectivePolicies(ctx context.Context, eps []ServiceEffectivePolicy) (int, error)
```
orishoshan, line 18: Since the appliers behave like reconcilers (they have all the same properties - supposed to bring the cluster to a desired state, idempotent, etc), let's use the same word (Reconcile?)

If it means the same thing it's better to use the same terminology

src/operator/effectivepolicy/Syncer.go
```diff
@@ -0,0 +1,142 @@
...
+	InjectRecorder(recorder record.EventRecorder)
+}
+
+type Syncer struct {
```
orishoshan, line 22: I would like to have a name that better communicates what this does. I get that it Syncs, but that's a rather generic word and this does a very specific thing.

ClusterwideEffectivePolicySyncer?

Or maybe (thinking out loud), let's think about what it does:
1. Calculates effective policies for all services in the cluster
2. Triggers "apply" (or reconcile if we take my other comment into account)

So perhaps it could be ReconcileAll? If the EffectivePoliciesAppliers turn into EffectivePoliciesReconcilers.

src/operator/effectivepolicy/Syncer.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/otterize/intents-operator/src/shared/errors"
+	"github.com/otterize/intents-operator/src/shared/injectablerecorder"
+	"github.com/otterize/intents-operator/src/shared/serviceidresolver/serviceidentity"
+	"github.com/samber/lo"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type EffectivePoliciesApplier interface {
```
orishoshan, line 17: This should be "effective policy applier" rather than "effective policies applier"

I forget what this concept is called in English grammar but here's an example to illustrate what I mean:
In English, do you say "train station" or "trains station", for a station where all trains are?
Similarly, this is an applier that handles effective policies, which makes it an effective policy applier.

An applier that handles trains would be a train applier, not a trains applier.

orishoshan (approved):

Issue #250, opened 2023-09-25
Add support for Linkerd authorization policies


orishoshan:
The intents operator supports reconciling ClientIntents into Istio authorization policies.

Proposal: Add support for Linkerd authorization policies, keeping with the same style.

orishoshan:
@aerosouund has taken this issue :)

orishoshan:
For this example ClientIntents:
```
apiVersion: k8s.otterize.com/v1alpha3
kind: ClientIntents
metadata:
  name: client
  namespace: otterize-tutorial-istio
spec:
  service:
    name: client
  calls:
    - name: nginx
      type: http
      HTTPResources:
        - path: /client-path
          methods: [ GET ]
```

1. Create a `Server` resource for `nginx` if it wasn't already created, as well as a `HTTPRoute` for the `HTTPResources` part (if there are any).
2. Create a `MeshTLSAuthentication` resource for `client` if it wasn't already created. For each client, create a separate `MeshTLSAuthentication` resource. I recommend using the name of the client in the name of the resource, such as `meshtls-client1`, `meshtls-client2`, and so on, to make it easy to find the resources when you need to update them or check if they already exist.
3. Create an `AuthorizationPolicy` resource that requires the `MeshTLSAuthentication` or `Server`.

As we discussed on Zoom, start with the `create` flow for `ClientIntents` and then move on to edit and delete. To understand the edit flow better, after creating a MeshTLS resource, scale down the intents operator to 0, then edit the ClientIntents so that the `client` is `client2`, and you should see two `MeshTLS` resources. In that situation the intents-operator needs to be able to detect that the first resource is now orphaned and delete it.

Sounds good @aerosouund?

aerosouund:
Sounds perfect, will move on with implementing this!

PR #324, opened 2023-12-30
Add support for Linkerd to the intents-operator

aerosouund:
### Description

Create required linkerd resources (servers, httproutes, authpolicies, mtls authentications, network authentications) based on otterize client intents.
An extra optional field has been added in the intent definition which is a port field.
It extends the same pattern used in istio policies but applies it to each respective linkerd resource

### References

- GitHub Issue/PR number addressed or fixed: https://github.com/otterize/intents-operator/issues/250

### Testing

There is a built image in a public repo that you can deploy in a cluster with linkerd installed: `aerosouund/intents-operator:0.1.34`

Also include details of the environment this PR was developed in (language/platform/browser version):
- golang 1.21
- kubernetes 1.23 on EKS
- Linkerd 2.x

### Checklist


aerosouund:
Over the upcoming period there's still some extra small functionality i want to add to the PR, but i've created it now to work on any large changes required at the same time

Including the following:
- don’t add the path in the name for the policies but validate if a policy should be created based on its targets, and add a random hash in the end of all policy names 
- cleaner implementation for deletion, instead of iterating over each type

- move the creation of primary resources to the start of the loop instead of checking if it should be done for each intent

- add support for grpc and tcp probes if they also break when a http route is created

- create a generateHTTPRouteName method

src/operator/config/crd/k8s.otterize.com_clientintents.patched
```diff
@@ -0,0 +1,285 @@
+apiVersion: apiextensions.k8s.io/v1
```
orishoshan, line 1: Please remove these generated .patched files from Git

src/operator/api/v1alpha3/clientintents_types.go
```diff
@@ -155,6 +157,9 @@ type Intent struct {
 	//+optional
 	Type IntentType `json:"type,omitempty" yaml:"type,omitempty"`
 
+	//+optional
+	Port int32 `json:"port,omitempty" yaml:"port,omitempty"`
```
orishoshan, line 161: We still need to figure out a way to do without specifying the port manually - let's try to figure this out when we talk

src/operator/controllers/intents_reconcilers/linkerd_reconciler.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	linkerdreconciler.linkerdManager = linkerdmanager.NewLinkerdManager(c, namespaces, &linkerdreconciler.InjectableRecorder, enforcementDefaultState, enableLinkerdPolicyCreation)
+	return linkerdreconciler
+}
+
+func (r *LinkerdReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logrus.Info("Starting linkerd reconcile logic")
+
+	installed, err := linkerdmanager.IsLinkerdServerInstalled(ctx, r.Client)
```
orishoshan, line 48:
```suggestion
	installed, err := linkerdmanager.IsLinkerdInstalled(ctx, r.Client)
```
This checks if the Linkerd Server CRD is installed, but in fact it's checking if Linkerd itself is installed in the cluster through this check.

src/operator/controllers/linkerd/linkerd_manager.go
```diff
@@ -0,0 +1,763 @@
...
+		Servers:               goset.NewSet[types.UID](),
+		AuthorizationPolicies: goset.NewSet[types.UID](),
+		Routes:                goset.NewSet[types.UID](),
+	}
+
+	for _, intent := range clientIntents.GetCallsList() {
+		if intent.Type != "" && intent.Type != otterizev1alpha3.IntentTypeHTTP && intent.Port != 0 { // this will skip non http ones, db for example, skip port doesnt exist as well
+			continue
+		}
+		shouldCreatePolicy, err := protected_services.IsServerEnforcementEnabledDueToProtectionOrDefaultState( //TODO:  check what that does
```
orishoshan, line 244: This is enabled through ProtectedService: https://docs.otterize.com/shadow-vs-active-enforcement#protected-services

src/operator/controllers/linkerd/linkerd_manager.go
```diff
@@ -0,0 +1,763 @@
...
+	clientServiceAccount string, // supplied in the create method
+) (map[string]*goset.Set[types.UID], error) {
+	currentResources := map[string]*goset.Set[types.UID]{
+		Servers:               goset.NewSet[types.UID](),
+		AuthorizationPolicies: goset.NewSet[types.UID](),
+		Routes:                goset.NewSet[types.UID](),
+	}
+
+	for _, intent := range clientIntents.GetCallsList() {
+		if intent.Type != "" && intent.Type != otterizev1alpha3.IntentTypeHTTP && intent.Port != 0 { // this will skip non http ones, db for example, skip port doesnt exist as well
```
orishoshan, line 241: The port condition seems inverted, as in if port is NOT empty AND type is not HTTP it will continue. But perhaps you meant

if intent.Type != "" // if intent type is not empty
&& intent.Type != IntentTypeHTTP // and intent type is not http
|| intent.Port == "" // OR port is empty

aquam8:
we'd be very interested in having support for Linkerd2 - then Otterize would really be a compelling solution for us as light user of EKS & Linkerd2 (mTLS, grpc load balancing but not leveraging authorization policies in linkerd). We use kubernetes network policies but we do not like it for the same reason you highlighted in [your blog post](https://otterize.com/blog/network-policies-are-not-the-right-abstraction). We're making a case towards dropping support for network policies all together (we are using Calico & aws vpc-cni) and we would be interesting in assessing how linkerd2/otterize could help with authorization policies on that side of the fence instead.

orishoshan:
That's great to hear @aquam8! We've been working on a couple of infrastructure changes that are starting to come in and have blocked this PR, namely #409 and #439, so Linkerd support should be getting merged soon!

aquam8:
Could this work be revised now that the dependent PRs have been merged, thank you!

PR #328, opened 2024-01-04
Support applying intents to API server "kubernetes" service, even when it doesn't have a pod

NetanelBollag:
### Description

It's a common implementation to have the control plane API server service with no pod, just an endpoint mapped to a server outside of the cluster network. Since egress enforcement requires explicitly enabling access to the API server, this PR allow the user to declare intents to the service in the format: `svc:kubernetes.default`, and the result will be an egress network policy to the server IP address and port.

### Testing

### Checklist

Egress network policy is an experimental feature, full documentation will be available on official release.

### DIFF
--- a/src/operator/controllers/intents_controller.go
+++ b/src/operator/controllers/intents_controller.go
@@ -193,6 +193,7 @@ func (r *IntentsReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&otterizev1alpha3.ClientIntents{}).
 		WithOptions(controller.Options{RecoverPanic: lo.ToPtr(true)}).
 		Watches(&otterizev1alpha3.ProtectedService{}, handler.EnqueueRequestsFromMapFunc(r.mapProtectedServiceToClientIntents)).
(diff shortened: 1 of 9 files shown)


src/operator/controllers/intents_controller.go
```diff
@@ -203,6 +204,32 @@ func (r *IntentsReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+func (r *IntentsReconciler) getIntentsToAPIServerService() []otterizev1alpha3.ClientIntents {
+	intentsToReconcile := make([]otterizev1alpha3.ClientIntents, 0)
+	fullServerName := fmt.Sprintf("svc:%s.%s", otterizev1alpha3.KubernetesAPIServerName, otterizev1alpha3.KubernetesAPIServerNamespace)
+	var intentsToServer otterizev1alpha3.ClientIntentsList
+	err := r.client.List(context.Background(),
+		&intentsToServer,
+		&client.MatchingFields{otterizev1alpha3.OtterizeTargetServerIndexField: fullServerName},
+	)
+	if err != nil {
```
amitlicht, line 224: you should return here.

amitlicht (approved):
see one comment inside

PR #335, opened 2024-01-10
Refactor `EgressNetworkPolicyReconciler` to reconcile `ServiceEffectivePolicies` instead of `ClientIntents`

omris94:
### Description

Following the [introduction](https://github.com/otterize/intents-operator/pull/323) of `serviceEffectivePolicy` reconciliation, refactor `EgressNetworkPolicyReconciler` to be `serviceEffectivePolicy` based

### References

https://github.com/otterize/intents-operator/pull/323

### Testing

The old tests were modified.

### Checklist


src/operator/controllers/intents_reconcilers/egress_network_policy/egress_network_policy.go
```diff
@@ -48,87 +50,82 @@ func NewEgressNetworkPolicyReconciler(
...
-			r.RecordWarningEventf(intents, consts.ReasonRemovingEgressNetworkPolicyFailed, "could not remove network policies: %s", err.Error())
-			return ctrl.Result{}, errors.Wrap(err)
-		}
-		return ctrl.Result{}, nil
-	}
+		ep.Service.Name, ep.Service.Namespace)
 
-	createdNetpols := 0
-	for _, intent := range intents.GetCallsList() {
+	touchedNetpols := make([]types.NamespacedName, 0)
```
omris94, line 61: rename because it's not just the "touched" network policies - it is all of them

src/operator/controllers/intents_reconcilers/egress_network_policy/egress_network_policy.go
```diff
@@ -48,87 +50,82 @@ func NewEgressNetworkPolicyReconciler(
...
-
-	if err != nil {
-		return ctrl.Result{}, errors.Wrap(err)
-	}
-
-	if intents.Spec == nil {
-		return ctrl.Result{}, nil
+func (r *EgressNetworkPolicyReconciler) applyEffectivePolicy(ctx context.Context, ep effectivepolicy.ServiceEffectivePolicy) ([]types.NamespacedName, error) {
+	if ep.ClientIntent == nil || ep.ClientIntent.Spec == nil {
+		return nil, nil
```
orishoshan, line 55:
```suggestion
		return make([]types.NamespacedName, 0), nil
```
Empty slice instead of nil slice. If err != nil then the other return value must always be initialized

src/operator/controllers/intents_reconcilers/egress_network_policy/egress_network_policy.go
```diff
@@ -48,87 +50,82 @@ func NewEgressNetworkPolicyReconciler(
...
-	}
-
-	if err != nil {
-		return ctrl.Result{}, errors.Wrap(err)
-	}
-
-	if intents.Spec == nil {
-		return ctrl.Result{}, nil
+func (r *EgressNetworkPolicyReconciler) applyEffectivePolicy(ctx context.Context, ep effectivepolicy.ServiceEffectivePolicy) ([]types.NamespacedName, error) {
+	if ep.ClientIntent == nil || ep.ClientIntent.Spec == nil {
```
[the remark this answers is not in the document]
orishoshan: If you are using this as a proxy to determine if we have any client intents and if not then we should have no egress netpols, this seems like something that should be on the ServiceEffectivePolicy itself, and not parsed here from the ClientIntents? That is, we have `CalledBy`, why not have `Calls`?

PR #339, opened 2024-01-17
Refactor  `EgressNetworkPolicyReconciler `, `InternetNetworkPolicyReconciler`, `PortNetworkPolicyReconciler` and `PortEgressNetworkPolicyReconciler` to reconcile `ServiceEffectivePolicy` instead of `ClientIntents`

omris94:
### Description

Following the [introduction](https://github.com/otterize/intents-operator/pull/323) of `serviceEffectivePolicy` reconciliation, refactor `EgressNetworkPolicyReconciler `, `InternetNetworkPolicyReconciler`, `PortNetworkPolicyReconciler` and `PortEgressNetworkPolicyReconciler` to be `serviceEffectivePolicy` based

### References

https://github.com/otterize/intents-operator/pull/323

### Testing

The old tests were modified.

### Checklist


src/operator/controllers/intents_reconcilers/port_egress_network_policy/port_egress_network_policy.go
```diff
@@ -51,38 +53,36 @@ func NewPortEgressNetworkPolicyReconciler(
 	}
 }
 
-func (r *PortEgressNetworkPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	intents := &otterizev1alpha3.ClientIntents{}
-	err := r.Get(ctx, req.NamespacedName, intents)
-	if k8serrors.IsNotFound(err) {
-		return ctrl.Result{}, nil
+func (r *PortEgressNetworkPolicyReconciler) ReconcileEffectivePolicies(ctx context.Context, eps []effectivepolicy.ServiceEffectivePolicy) (int, error) {
+	currentPolicies := goset.NewSet[types.NamespacedName]()
```
orishoshan, line 57: Is this only used as a counter?

orishoshan: Because this is a set, if you accidentally add the same policy twice, then it will only appear once. But if you did add it twice due to a bug, I think it would cause the policy to have the wrong data, and the set would make it harder to figure out that this is the case (that it was handled twice).

orishoshan: Consider using a slice instead

src/operator/effectivepolicy/groupreconciler.go
```diff
@@ -73,10 +73,16 @@ func (g *GroupReconciler) getAllServiceEffectivePolicies(ctx context.Context) ([
...
 	services := goset.NewSet[serviceidentity.ServiceIdentity]()
 	for _, clientIntent := range intentsList.Items {
+		if !clientIntent.DeletionTimestamp.IsZero() {
+			continue
+		}
 		service := serviceidentity.ServiceIdentity{Name: clientIntent.Spec.Service.Name, Namespace: clientIntent.Namespace}
 		services.Add(service)
 		serviceToIntent[service] = clientIntent
 		for _, intentCall := range clientIntent.GetCallsList() {
+			if intentCall.IsTargetServerKubernetesService() {
```
orishoshan, line 83: This would mean that the service is not counted as a server, right?
So how are ingress policies created for it?

src/operator/controllers/intents_reconcilers/port_egress_network_policy/port_egress_network_policy.go
```diff
@@ -96,90 +96,87 @@ func (r *PortEgressNetworkPolicyReconciler) Reconcile(ctx context.Context, req c
...
-		createdPolicies, err := r.handleNetworkPolicyCreation(ctx, intents, intent, req.Namespace)
+		policy, err := r.handleNetworkPolicyCreation(ctx, intents, intent, intents.Namespace)
 		if err != nil {
 			r.RecordWarningEventf(intents, consts.ReasonCreatingEgressNetworkPoliciesFailed, "could not create network policies: %s", err.Error())
-			return ctrl.Result{}, errors.Wrap(err)
+			return nil, errors.Wrap(err)
 		}
-		if createdPolicies {
-			createdNetpols += 1
+		if policy != nil {
```
orishoshan, line 104: When is policy nil?
This should never happen - if err == nil then the other param should never be nil.

src/operator/effectivepolicy/groupreconciler.go
omris94: They are created using the same method as the egress ones. This is something we can change when we start using existing pod selectors instead of labeling them ourselves.

src/operator/controllers/intents_reconcilers/port_egress_network_policy/port_egress_network_policy.go
```diff
@@ -51,135 +53,127 @@ func NewPortEgressNetworkPolicyReconciler(
...
+	if len(errorList) > 0 {
+		return 0, errors.Wrap(goerrors.Join(errorList...))
 	}
 
-	if intents.Spec == nil {
-		return ctrl.Result{}, nil
-	}
+	// remove policies that doesn't exist in the policy list
+	err := r.removeNetworkPoliciesThatShouldNotExist(ctx, currentPolicies)
+	return currentPolicies.Len(), err
```
orishoshan, line 73:
```suggestion
	return currentPolicies.Len(), errors.Wrap(err)
```

src/operator/effectivepolicy/groupreconciler.go
```diff
@@ -95,35 +97,59 @@ func (g *GroupReconciler) getAllServiceEffectivePolicies(ctx context.Context) ([
...
+			ep.Calls = append(ep.Calls, clientIntents.GetCallsList()...)
+			ep.ClientIntentsEventRecorder = injectablerecorder.NewObjectEventRecorder(&g.InjectableRecorder, lo.ToPtr(clientIntents))
 		}
 		epSlice = append(epSlice, ep)
 	}
 
 	return epSlice, nil
 }
 
+// a function that checks if we should create a SEP for a given intent target server
```
orishoshan, line 111: Is the comment necessary? Seems to be saying the same thing as the function's name and is likely to go out of date when refactored :)
If you absolutely need to keep it, consider:
```suggestion
// shouldCreateEffectivePolicyForIntentTargetServer that checks if we should create a SEP for a given intent target server
```
So that IDEs will find this line as well and hopefully the person refactoring will notice they need to update the comment.

src/operator/effectivepolicy/groupreconciler.go
```diff
@@ -95,35 +97,59 @@ func (g *GroupReconciler) getAllServiceEffectivePolicies(ctx context.Context) ([
...
+func (g *GroupReconciler) shouldCreateEffectivePolicyForIntentTargetServer(intent v1alpha3.Intent) bool {
+	if intent.IsTargetOutOfCluster() {
+		return false
+	}
+	if intent.IsTargetServerKubernetesService() {
```
orishoshan, line 116:
```suggestion
        // Services are currently unused when used as a target, since the policy is created by looking at client Calls.
	if intent.IsTargetServerKubernetesService() {
```

orishoshan (approved):

PR #345, opened 2024-01-23
Add a basic e2e test for egress network policy creation.

omris94:
### Description

Add a basic e2e test for egress network policy creation.

### Testing

### Checklist

### DIFF
--- a/.github/workflows/netpol-e2e-test.yaml
+++ b/.github/workflows/netpol-e2e-test.yaml
@@ -89,6 +89,9 @@ jobs:
       - name: Wait for Otterize
         run: |-
           kubectl wait pods -n otterize-system -l app=intents-operator --for condition=Ready --timeout=360s
+          # wait for webhook to be ready
+          POD_IP=`kubectl get pod -l app=intents-operator -n otterize-system -o=jsonpath='{.items[0].status.podIP}'`
+          kubectl wait -n otterize-system --for=jsonpath='{.subsets[0].addresses[0].ip}'=$POD_IP endpoints/intents-operator-webhook-service
 
 
       - name: Wait for Tutorial services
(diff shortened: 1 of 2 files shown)


.github/workflows/netpol-e2e-test.yaml
```diff
@@ -89,6 +89,9 @@ jobs:
       - name: Wait for Otterize
         run: |-
           kubectl wait pods -n otterize-system -l app=intents-operator --for condition=Ready --timeout=360s
+          # wait for webhook to be ready
```
NetanelBollag, line 92: Oh yes thank you for that!!!

.github/workflows/netpol-e2e-test.yaml
```diff
@@ -230,10 +237,133 @@ jobs:
...
+          
+          # should not work at first because there is no allow DNS netpol
+          echo check client log
+          wait_for_log $CLI1_POD 10 "curl timed out"
+          
+          # should be blocked (using 3 because the log should repeat itself every 3 lines)
+          echo check client other log
+
+          wait_for_log $CLI2_POD 10 "curl timed out"
+
```
NetanelBollag, line 347: maybe add echo for every step?

omris94: I'll do it

.github/workflows/netpol-e2e-test.yaml
```diff
@@ -230,10 +237,133 @@ jobs:
...
+          source .github/workflows/test-bashrc.sh
+          
+          kubectl apply -f https://docs.otterize.com/code-examples/automate-network-policies/intents.yaml
+          
+          # should not work at first because there is no allow DNS netpol
+          echo check client log
+          wait_for_log $CLI1_POD 10 "curl timed out"
+          
+          # should be blocked (using 3 because the log should repeat itself every 3 lines)
+          echo check client other log
```
NetanelBollag, line 344: IMHO it's better:
```suggestion
          echo "check client other log"
```

.github/workflows/netpol-e2e-test.yaml
```diff
@@ -230,10 +237,133 @@ jobs:
...
+           # should work because there is an applied intent and allowDNS netpol
+          echo check client log
+          wait_for_log $CLI1_POD 10 "Hi, I am the server, you called, may I help you?"
+          
+           # should be blocked (using 3 because the log should repeat itself every 3 lines)
+          echo check client other log
+
+          wait_for_log $CLI2_POD 10 "curl timed out"
+  
+
```
NetanelBollag, line 359: This test is actually checking nothing is totally broken when egress enabled, but I suggest there will be another service, `server2` that only `client` will call (won't be called by `client-other`). When intents applied we should expect communication with `server2` to stop, and would assert that egress policy actually work with blocking the *client* un-declared calls to servers

NetanelBollag (commented):
Generally look good, the only thing is missing IMHO is to actually test egress logic itself

PR #364, opened 2024-02-28
Add periodic reconcile for database intents to include newly created tables.

evyatarmeged:
### Description
This is relevant for database intents without a specified table or with a `*` in the table field.


src/operator/controllers/intents_controller.go
```diff
@@ -116,6 +117,7 @@ func NewIntentsReconciler(
 	if enforcementConfig.EnableDatabasePolicy {
 		databaseReconciler := database.NewDatabaseReconciler(client, scheme, otterizeClient)
 		intentsReconciler.group.AddToGroup(databaseReconciler)
+		intentsReconciler.DatabaseReconciler = databaseReconciler
```
orishoshan, line 120: It's bad practice to change an object from outside. But I'm not sure why this needs to go in the intents reconciler? The database reconciler should handle this all on its own.

PR #384, opened 2024-03-11
Bugfix: avoid splitting target server name on intents of type Azure (and other non-k8s types)

amitlicht:
### Description

In intents of type Azure (and other non-k8s types), we should never try to split the server "name" to name & namespace. 
This bug breaks intents with Azure resource names such as `/providers/Microsoft.Storage/storageAccounts/otterizetutorialazureiam/blobServices/default/containers/test`, as the name is split around the period when reporting to Otterize cloud. 

Additionally, fixed a bug where if the intents operator attempts to handle a deleted `ClientIntents` resource whose workload identity no longer exists, it fails to reconcile it. 

### Checklist

### DIFF
--- a/src/operator/api/v1alpha3/clientintents_types.go
+++ b/src/operator/api/v1alpha3/clientintents_types.go
@@ -323,6 +323,10 @@ func (in *Intent) GetTargetServerName() string {
 		return OtterizeInternetTargetName
 	}
 
+	if in.Type == IntentTypeAWS || in.Type == IntentTypeGCP || in.Type == IntentTypeAzure || in.Type == IntentTypeDatabase {
+		return in.Name
+	}
+
 	if in.IsTargetServerKubernetesService() {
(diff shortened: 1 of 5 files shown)


src/operator/api/v1alpha3/clientintents_types.go
```diff
@@ -323,6 +323,10 @@ func (in *Intent) GetTargetServerName() string {
 		return OtterizeInternetTargetName
 	}
 
+	if in.Type == IntentTypeAWS || in.Type == IntentTypeGCP || in.Type == IntentTypeAzure || in.Type == IntentTypeDatabase {
```
orishoshan, line 326: 👍 
FYI: Once @omris94's changes go in, we should be able to replace this with a generic condition (either in.Kind == "", or if we end up settling on a schema that has a different way to differentiate internal from external targets)

orishoshan (approved):

PR #392, opened 2024-03-18
Database enforcement support - new CRD and database permission configuration 

evyatarmeged:
### Description
Adding support for configuring database permissions from the intents operator. Main highlights:
* New `PostgreSQLServerConfig` CRD to describe Postgres servers - their address & credentials
* Added logic for configuring permissions as part of the database reconciler
* Added persistent cluster ID configmap in the Otterize namespace so ensure uniqueness in shared databases between similarly deployed clusters
* New validation webhook to enforce cluster wide name uniqueness of PGServerConf even though it is namespace scoped
* Tests to follow   

### Testing

### Checklist


src/operator/api/v1alpha3/postgresqlserverconfig_types.go
```diff
@@ -0,0 +1,72 @@
...
+type DBPermissionChange string
+
+const (
+	DBPermissionChangeApply  DBPermissionChange = "APPLY"
+	DBPermissionChangeDelete DBPermissionChange = "DELETE"
+)
+
+type DatabaseCredentials struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
```
amitlicht, line 32: Does this mean we store the password plaintext on the CRD? Seems like bad practice.

src/operator/config/crd/k8s.otterize.com_clientintents.patched
```diff
@@ -2,7 +2,8 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.14.0
+    controller-gen.kubebuilder.io/version: v0.9.0
```
amitlicht, line 5: These diffs indicate you ran something weird in your make process. Been there. It's not exactly well documented 🤦‍♀️ 
try: 
```
make copy-manifests-to-helm
```

src/operator/controllers/intents_reconcilers/database/database.go
```diff
@@ -39,51 +42,99 @@ func NewDatabaseReconciler(
...
 			continue
 		}
-
-		intentInput := intent.ConvertToCloudFormat(intents.Namespace, intents.GetServiceName())
-		intentInputList = append(intentInputList, intentInput)
+		dbIntents = append(dbIntents, intent)
 	}
 
-	if len(intentInputList) == 0 {
+	if len(dbIntents) == 0 {
```
amitlicht, line 73: What happens here if a ClientIntent was modified (not deleted), and all of its DB intents were removed? Shouldn't we still reconcile _something_, to remove any existing permissions given to the client?

src/operator/controllers/intents_reconcilers/database/database.go
```diff
@@ -39,51 +42,99 @@ func NewDatabaseReconciler(
...
-		errType, errMsg, ok := graphqlclient.GetGraphQLUserError(err)
-		if !ok || errType != graphqlclient.UserErrorTypeAppliedIntentsError {
-			r.RecordWarningEventf(intents, ReasonApplyingDatabaseIntentsFailed, "Failed applying database intents: %s", err.Error())
+	databaseToIntents, err := r.MapDBInstanceToIntents(dbIntents)
+	if err != nil {
+		r.RecordWarningEventf(clientIntents, ReasonApplyingDatabaseIntentsFailed, "Failed applying database clientIntents: %s", err.Error())
+		return ctrl.Result{}, errors.Wrap(err)
+	}
+
+	for databaseInstance, intents := range databaseToIntents {
```
amitlicht, line 83: Same here. How do we handle database instances which had prior intents, and now, following the modified intent, have 0 intents? We should still remove any existing permissions.

src/operator/controllers/intents_reconcilers/database/database.go
```diff
@@ -39,51 +42,99 @@ func NewDatabaseReconciler(
...
+		pgServerConfigs := otterizev1alpha3.PostgreSQLServerConfigList{}
+		err := r.client.List(ctx, &pgServerConfigs)
+		if err != nil {
+			r.RecordWarningEventf(clientIntents, ReasonErrorFetchingPostgresServerConfig,
+				"Error trying to fetch '%s' PostgresServerConf for client '%s'. Error: %s",
+				databaseInstance, clientIntents.GetServiceName(), err.Error())
+			return ctrl.Result{}, nil
```
amitlicht, line 90: Why is this returning `nil` and not the error?

src/operator/controllers/intents_reconcilers/database/database.go
```diff
@@ -39,51 +42,99 @@ func NewDatabaseReconciler(
...
+		}
+		pgServerConf, err := findMatchingPGServerConfForDBInstance(databaseInstance, pgServerConfigs)
+		if err != nil {
+			r.RecordWarningEventf(clientIntents, ReasonMissingPostgresServerConfig,
+				"Could not find matching PostgreSQLServerConfig. Error: %s", err.Error())
+			return ctrl.Result{}, nil
```
amitlicht, line 96: same

src/operator/controllers/intents_reconcilers/database/database.go
```diff
@@ -39,51 +42,99 @@ func NewDatabaseReconciler(
...
 	}
 
-	r.RecordNormalEventf(intents, ReasonAppliedDatabaseIntents, "Database intents reconcile complete, reconciled %d intent calls", len(intentInputList))
+	r.RecordNormalEventf(clientIntents, ReasonAppliedDatabaseIntents, "Database clientIntents reconcile complete, reconciled %d intent calls", len(dbIntents))
 
 	return ctrl.Result{}, nil
 }
+
+func (r *DatabaseReconciler) MapDBInstanceToIntents(intents []otterizev1alpha3.Intent) (map[string][]otterizev1alpha3.Intent, error) {
+	dbInstanceToIntents := map[string][]otterizev1alpha3.Intent{}
```
amitlicht, line 114: `lo.GroupBy` does that nicely for you.

src/operator/controllers/intents_reconcilers/database/database.go
```diff
@@ -39,51 +42,99 @@ func NewDatabaseReconciler(
...
+	}
+
+	return dbInstanceToIntents, nil
+}
+
+func findMatchingPGServerConfForDBInstance(
+	databaseInstanceName string,
+	pgServerConfigList otterizev1alpha3.PostgreSQLServerConfigList) (*otterizev1alpha3.PostgreSQLServerConfig, error) {
+
+	for _, conf := range pgServerConfigList.Items {
```
amitlicht, line 132: can use `lo.Find`

amitlicht: I know I'm considered an lo-enthusiastic in the team, it's just that it removes so much unnecessary code 😍

src/operator/databaseconfigurator/db_helpers.go
```diff
@@ -0,0 +1,23 @@
...
+	"fmt"
+	"strings"
+)
+
+const PGUsernameMaxLength = 63
+
+func BuildPostgresUsername(clusterUID string, workloadName, namespace string) string {
+	username := fmt.Sprintf("%s_%s_%s", KubernetesToPostgresName(workloadName), KubernetesToPostgresName(namespace), KubernetesToPostgresName(clusterUID))
+	if len(username) > PGUsernameMaxLength {
+		username = username[:PGUsernameMaxLength]
```
amitlicht, line 13: You might be trimming the clusterUID here, which means that your name is going to suffer from uniqueness issues. 
This didn't use to happen when this code ran in Otterize Cloud, because the first chunk of the name was the otterize service ID, which guaranteed uniqueness on its own. This is not the case anymore, and so you should make sure you add a unique id, or suffix the trimmed name with a hash over the full username (we use a similar method in the IAM policy agents, look for `TruncateHashName`). 

Also, I can already spot a potential issue with Postgres visibility here. We use the postgres username in postgres visibility to detect to which service this traffic belongs. We need to make sure we somehow preserve this functionality even when the usernames are generated by the credentials operator.

src/operator/databaseconfigurator/postgres.go
```diff
@@ -0,0 +1,528 @@
+package databaseconfigurator
```
amitlicht, line 1: This package will be better located under `src/operator/controller/intents_reconcilers/database`

amitlicht: Did not review this file because I guess it was copied almost as-is from otterize/cloud. I will say it needs some love and attention, but leaving that out of scope for now.

src/operator/webhooks/postgresqlserverconfigs_webhook.go
```diff
@@ -0,0 +1,119 @@
...
+			return &field.Error{
+				Type:     field.ErrorTypeDuplicate,
+				Field:    "name",
+				BadValue: currName,
+				Detail: fmt.Sprintf(
+					"Postgres server config already exists with name %s. Existing resource: %s.%s",
+					currName, conf.Name, conf.Namespace),
+			}
+		}
+	}
```
amitlicht, line 117: can use lo.Find here too

amitlicht (changes requested):

src/operator/api/v1alpha3/postgresqlserverconfig_types.go
evyatarmeged: Going to be changed to a different CRD that holds credentials (and should have similar access permissions)

src/operator/controllers/intents_reconcilers/database/database.go
evyatarmeged: We don't want to spam reconcile errors since retrying won't work (this is pending a user action - create the missing CRD). We're failing gracefully and posting an event on the intents object.
When the user creates a PGServerConf the intents reconciler will be called again (it has a `.Watches` on the pg serverconf)

src/shared/local.env
```diff
@@ -1,20 +0,0 @@
-# This file is used for local development and testing of the operator
```
amitlicht, line 1: see my comment re the gitignore file.

amitlicht (changes requested):

src/operator/controllers/intents_controller.go
```diff
@@ -226,6 +227,40 @@ func (r *IntentsReconciler) mapProtectedServiceToClientIntents(ctx context.Conte
...
+	err := r.client.List(context.Background(),
+		&intentsList,
+		&client.MatchingFields{otterizev1alpha3.OtterizeTargetServerIndexField: dbInstanceName},
+	)
+	if err != nil {
+		logrus.Errorf("Failed to list client intents targeting %s: %v", dbInstanceName, err)
+	}
+
+	intentsToReconcile = append(intentsToReconcile, intentsList.Items...)
+	return intentsToReconcile
```
amitlicht, line 261: you can just return `intentsList.Items`,  there's never anything else in `intentsToReconcile`.

src/operator/controllers/intents_reconcilers/database/database_reconciler.go
```diff
@@ -0,0 +1,159 @@
...
+			return ctrl.Result{}, nil // Not returning error on purpose, missing PGServerConf - record event and move on
+		}
+		pgConfigurator := postgres.NewPostgresConfigurator(pgServerConf.Spec, r.client)
+		err = pgConfigurator.ConfigureDBFromIntents(ctx, clientIntents.GetServiceName(), clientIntents.Namespace, intents, action)
+		if err != nil {
+			r.RecordWarningEventf(clientIntents, ReasonApplyingDatabaseIntentsFailed,
+				"Failed applying database clientIntents: %s", err.Error())
+			return ctrl.Result{}, errors.Wrap(err)
+		}
+	}
```
amitlicht, line 93: [Optional] Consider parallelizing this across DB instances (using errgroup). This comment applies in other places where you iterate over different DB instances.

src/shared/clusterutils/username.go
```diff
@@ -0,0 +1,47 @@
...
+	if len(username) > HashedUsernameSectionMaxLength {
+		username = username[:HashedUsernameSectionMaxLength]
+	}
+
+	if len(namespace) > HashedUsernameSectionMaxLength {
+		namespace = namespace[:HashedUsernameSectionMaxLength]
+	}
+
+	if len(clusterUID) > HashedUsernameSectionMaxLength {
+		clusterUID = clusterUID[:HashedUsernameSectionMaxLength]
```
amitlicht, line 29: well, the linter has a point. There's some mess left in this function. 
Consider adding a UT for it to catch this mess.

src/shared/databaseconfigurator/postgres/postgres.go
```diff
@@ -0,0 +1,480 @@
...
+		return "", errors.Errorf("received sanitize input '%v' with type '%T', which was not string or DatabaseOperation", formatInput, formatInput)
+	}
+	return fmt.Sprintf(string(s), sanitizedItems...), nil
+}
+
+type PostgresConfigurator struct {
+	conn         *pgx.Conn
+	databaseInfo otterizev1alpha3.PostgreSQLServerConfigSpec
+	setConnMutex sync.Mutex
+	client       client.Client
```
amitlicht, line 81: I'm pretty sure you're not using this member.

amitlicht (approved):

PR #394, opened 2024-03-19
Report services that won't be blocked due to external access even when no network policy

NetanelBollag:
### Description

Support showing services that won't be blocked due to external access policy in shadow mode

### Testing

- [x] This change adds test coverage for new functionality

### Checklist

### DIFF
--- a/src/operator/controllers/external_traffic/shared.go
+++ b/src/operator/controllers/external_traffic/shared.go
@@ -1,8 +1,15 @@
 package external_traffic
 
 import (
+	"context"
+	"fmt"
+	"github.com/otterize/intents-operator/src/operator/api/v1alpha3"
+	"github.com/otterize/intents-operator/src/shared/errors"
+	"github.com/otterize/intents-operator/src/shared/otterizecloud/graphqlclient"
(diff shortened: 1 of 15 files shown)


src/operator/controllers/external_traffic/shared.go
```diff
@@ -23,3 +30,51 @@ func serviceNamesFromIngress(ingress *v1.Ingress) sets.Set[string] {
...
+	case corev1.ServiceTypeLoadBalancer:
+		cloudServiceType = graphqlclient.KubernetesServiceTypeLoadBalancer
+	case corev1.ServiceTypeNodePort:
+		cloudServiceType = graphqlclient.KubernetesServiceTypeNodePort
+	case corev1.ServiceTypeClusterIP:
+		cloudServiceType = graphqlclient.KubernetesServiceTypeClusterIp
+	case corev1.ServiceTypeExternalName:
+		cloudServiceType = graphqlclient.KubernetesServiceTypeExternalName
+	default:
+		// Should not reach here until k8s adds new service type
```
amitlicht, line 55: consider reporting an error here, just to be on the safe side

amitlicht (approved):

PR #395, opened 2024-03-21
Move dedicated intents- and credentials- related parts of IAM agents out of shared and into operators

amitlicht:
### Description
This PR refactors some specific intents- and credentials- operator code out of the "shared" library and into the specific operators codebase. 
This simplifies the development workflow for both operators. 
Moreover, this PR introduces separate reconcile loops for every IAM agents, so that they will be able to succeed or fail regardless of if other IAM agents failed. 

### References
https://github.com/otterize/intents-operator/pull/395
https://github.com/otterize/credentials-operator/pull/123
https://github.com/otterize/helm-charts/pull/193

### Checklist


src/operator/controllers/intents_reconcilers/iam/iam_reconciler.go
```diff
@@ -17,26 +18,19 @@ import (
...
-	DeleteRolePolicyFromIntents(ctx context.Context, intents otterizev1alpha3.ClientIntents) error
-}
-
 type IAMIntentsReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 	injectablerecorder.InjectableRecorder
 	serviceIdResolver serviceidresolver.ServiceResolver
-	agents            []IAMPolicyAgent
+	agents            []iampolicyagents.IAMPolicyAgent
```
orishoshan, line 26: Overall looks good, my feedback on this bit remains: we should have multiple reconcilers, one per "agent" (essentially independent operations should be in separate reconcilers, so they don't create a dependency).

At KubeCon we had a case where the AWS intents agent didn't work and as a result the Google Cloud intents agent didn't run, since the AWS intents agent ran before it and failed.

This is solvable by queueing requests separately for each agent and handling/retrying errors separately - and that is what the Reconciler abstraction from controller-runtime does.

So my suggestion is: we should have 3 IAMIntentsReconcilers, each registered separately, and they each should take a single IAMPolicyAgent.

amitlicht:
@orishoshan I applied your feedback, we now have multiple reconcilers, one for IAM agent type. 
Would appreciate your re-review.

orishoshan (approved):

PR #401, opened 2024-04-02
Group bugsnag error reports by class, file & method name (ignoring line numbers)

amitlicht:
### Description

By default, bugsnag groups errors sharing the same error class, file, and line number of the top in-project stack frame of the innermost exception.
We override this to group by error class, file, and *method name* (without line number) of the top in-project stack frame.
This should result in reducing alert noise and avoid trigerring new alerts when the line number issuing the alert is changed (e.g. because of changes in other places in the same file). 

### References
- https://docs.bugsnag.com/product/error-grouping/#top-in-project-stack-frame

### Checklist

### DIFF
--- a/src/shared/telemetries/errorreporter/errorreporter.go
+++ b/src/shared/telemetries/errorreporter/errorreporter.go
@@ -2,12 +2,14 @@ package errorreporter
 
 import (
 	"context"
+	"fmt"
 	"github.com/bugsnag/bugsnag-go/v2"
 	"github.com/otterize/intents-operator/src/shared/errors"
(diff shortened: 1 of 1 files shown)


NetanelBollag:
I want to add the ability to pass an error that is unique based on a unique value that can be set when calling `errors.Wrap(err)`. This will allow us to make errors returning from cloud or internal api unique based on the error code. 
For example:
```
res, err := otterizeclient.ReportStatus()
if err != nil {
  return errors.WrapUnique(err, res.HTTPStatus)
}
```

This can be use by passing the HTTP status or GQL error to the wrap function, an to use it to set the error class which will also make it appear on Busnag UX.

PR #417, opened 2024-04-21
Fix error reporting for errors caused by sentinel errors

otterobert:
### Description

When creating a new error using bugsnag constructor it will freeze the error traceback. instead we fallback to go default error module that does not apply a traceback and bugsnag will add the traceback on throw.

### DIFF
--- a/src/operator/controllers/intents_reconcilers/networkpolicy/reconciler.go
+++ b/src/operator/controllers/intents_reconcilers/networkpolicy/reconciler.go
@@ -420,6 +420,9 @@ func (r *Reconciler) removeNetworkPolicy(ctx context.Context, networkPolicy v1.N
 		return errors.Wrap(err)
 	}
 	err = r.Delete(ctx, &networkPolicy)
+	if k8serrors.IsNotFound(err) {
+		return nil
+	}
 	if err != nil {
 		return errors.Wrap(err)
 	}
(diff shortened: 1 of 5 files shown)


src/shared/errors/errors.go
```diff
@@ -10,6 +10,10 @@ func New(text string) error {
 	return bugsnagerrors.New(text, 1)
 }
 
+func NewCustomError(text string) error {
```
orishoshan, line 13:
```suggestion
func NewSentinelError(text string) error {
```
Align terminology with common terminology:
https://go.dev/blog/go1.13-errors#:~:text=Sometimes%20we%20compare%20an%20error%20to%20a%20known%20sentinel%20value

src/operator/controllers/intents_reconcilers/networkpolicy/reconciler.go
```diff
@@ -420,6 +420,9 @@ func (r *Reconciler) removeNetworkPolicy(ctx context.Context, networkPolicy v1.N
 		return errors.Wrap(err)
 	}
 	err = r.Delete(ctx, &networkPolicy)
+	if k8serrors.IsNotFound(err) {
+		return nil
+	}
 	if err != nil {
```
orishoshan, line 426:
```suggestion
	if k8serrors.IgnoreNotFound(err) != nil {
```

orishoshan (approved):

PR #421, opened 2024-05-01
Addressing empty version before determining release stage

evyatarmeged:
### Description
Addressing empty version before determining release stage to avoid cases where locally built images with no version that happen to error, will get reported as production envs


src/shared/telemetries/errorreporter/errorreporter.go
```diff
@@ -71,6 +71,10 @@ func Init(componentName string, version string, apiKey string) {
 	logrus.AddHook(hook)
 }
 
+func isStagingVersion(version string) bool {
```
orishoshan, line 74:
```suggestion
func isStagingOrDevVersion(version string) bool {
```

orishoshan (approved):

PR #423, opened 2024-05-06
MySQL integration ClientIntents support

amitlicht:
### Description
- Add MySQLServerConfig CRD & webhooks
- Add DatabaseConfigurator interface, implemented by PostgresConfigurator and by MySQLConfigurator
- Support MySQLServerConfigs in intents reconcilers
- Move DB user creation to intents-operator scope, leaving the credentials-operator responsible only for password maintenance 

### References
Credentials-operator PR: https://github.com/otterize/credentials-operator/pull/134

### Testing

### Checklist


src/operator/controllers/intents_reconcilers/database/database_reconciler.go
```diff
@@ -169,6 +237,31 @@ func (r *DatabaseReconciler) getClusterID(ctx context.Context) (string, error) {
 	return clusterID, nil
 }
 
+func (r *DatabaseReconciler) annotateDatabaseAndUsernameOnPod(ctx context.Context, intents otterizev1alpha3.ClientIntents, username string, dbInstance string) error {
+	pod, err := r.serviceIdResolver.ResolveClientIntentToPod(ctx, intents)
```
amitlicht, line 241: what about when a pod goes up after the intents already exist? will this work?

amitlicht: another thought - what if there's more than one pod matching the clientintents? We will only annotate the first one we found. 

In general I think we should consider going the other way about it - using the intent-operator's pods reconciler, for every pod that has clientintents of type DB, annotate it with the db user & password. I.e., do this in the pod reconciler loop, and not in the intents reconciler loop. WDYT?

amitlicht: another thought - what if there's more than one pod matching the clientintents? We will only annotate the first one we found. 

In general I think we should consider going the other way about it - using the intent-operator's pods reconciler, for every pod that has clientintents of type DB, annotate it with the db user & password. I.e., do this in the pod reconciler loop, and not in the intents reconciler loop. @evyatarmeged WDYT?

src/operator/controllers/intents_reconcilers/database/database_reconciler.go
```diff
@@ -76,87 +89,150 @@ func (r *DatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+
+	missingDBServerConfig := lo.Without(lo.Keys(dbInstanceToIntents), existingDBInstances...)
+	for _, missingDBInstance := range missingDBServerConfig {
+		// Not returning error on purpose, missing PGServerConf - record event and move on
+		// When a new PGServerConf is created, the clientIntents will be reconciled again
+		r.RecordWarningEventf(clientIntents, ReasonMissingDBServerConfig,
+			"Missing database server config: did not find DB server config to match database '%s' in the cluster", missingDBInstance)
+	}
+
+	for _, config := range pgServerConfigs.Items {
```
evyatarmeged, line 116: Wouldn't missing server configs still be iterated here, causing an error ?

src/operator/controllers/intents_reconcilers/database/database_reconciler.go
```diff
@@ -76,87 +89,150 @@ func (r *DatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+		if err != nil {
+			r.RecordWarningEventf(clientIntents, ReasonApplyingDatabaseIntentsFailed,
+				"Failed applying database clientIntents: %s", err.Error())
+			return ctrl.Result{}, errors.Wrap(err)
 		}
 	}
 
-	if err := r.cleanExcessPermissions(ctx, pgUsername, clientIntents, pgServerConfigs); err != nil {
-		return ctrl.Result{}, errors.Wrap(err)
+	for _, config := range mySQLServerConfigs.Items {
```
evyatarmeged, line 125: Same question about missing server confs

src/operator/controllers/intents_reconcilers/database/database_reconciler.go
evyatarmeged: I think we should just log a warning in case no pods exist, and call the intents reconciler from the pods reconciler (like we do in `IAMPodReconciler`) for pods with DB intents. At least until it all moves to SEP.
@amitlicht thoughts ?

src/operator/controllers/intents_reconcilers/database/database_reconciler.go
amitlicht: no. 
`missingDBServerConfig` contains DB names referred to by intents, where the db name doesn't have a matching postgres/mysqlServerConfig. We won't iterate over them because they're not in `pgServerConfigs` / `mySQLServerConfigs`. 
We will (correctly) iterate over dbServerConfigs with no intents, to properly trigger cleanup of instances with no applied intents.

amitlicht: I'll rename some vars to make it clearer.

evyatarmeged (approved):

PR #425, opened 2024-05-07
Fixed bug where externally accessible services would be reported with an incorrect name sometimes, causing them to be incorrectly taken into account in the access graph simulation

orishoshan:
TODO: Handle new pods starting up, in the case that the service has not been reported because no pods were up.

Please see the [contributing guidelines](CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo. This template is based on Auth0's excellent template.

### Description

Describe the purpose of this PR along with any background information and the impacts of the proposed change. For the benefit of the community, please do not assume prior context.

Provide details that support your chosen implementation, including: breaking changes, alternatives considered, changes to the API, etc.

### References

Include any links supporting this change such as a:

- GitHub Issue/PR number addressed or fixed
- StackOverflow post
- Related pull requests/issues from other repos

If there are no references, simply delete this section.

### Testing

### Checklist

### DIFF
--- a/src/operator/controllers/external_traffic/service_uploader.go
+++ b/src/operator/controllers/external_traffic/service_uploader.go
@@ -2,16 +2,27 @@ package external_traffic
 
 import (
 	"context"
+	"fmt"
+	lru "github.com/hashicorp/golang-lru"
+	"github.com/otterize/intents-operator/src/operator/api/v1alpha3"
+	"github.com/otterize/intents-operator/src/shared"
 	"github.com/otterize/intents-operator/src/shared/errors"
 	"github.com/otterize/intents-operator/src/shared/operator_cloud_client"
 	"github.com/otterize/intents-operator/src/shared/otterizecloud/graphqlclient"
+	"github.com/otterize/intents-operator/src/shared/serviceidresolver"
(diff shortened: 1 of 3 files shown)


src/operator/controllers/external_traffic/service_uploader.go
```diff
@@ -53,22 +83,92 @@ func (s *ServiceUploaderImpl) UploadNamespaceServices(ctx context.Context, names
...
+	s.namespaceToReportedServicesCache.Add(namespace, hashSum)
+
 	return nil
 }
 
-func (s *ServiceUploaderImpl) getExternalService(ctx context.Context, svc *corev1.Service) (graphqlclient.ExternallyAccessibleServiceInput, bool, error) {
+func reportedServicesCacheValuePart(namespace, serverName string) []byte {
+	return []byte(fmt.Sprintf("%s%s", serverName, namespace))
+}
+
```
NetanelBollag, line 94: I think it's better to print here the whole struct, to make sure any change in any field, also on future updates, will cause upload.
```suggestion

func reportedServicesCacheValuePart(service graphqlclient.ExternallyAccessibleServiceInput) []byte {
	return []byte(fmt.Sprintf("\"%+v", service))
}

```

NetanelBollag (approved):

PR #435, opened 2024-05-26
Use new error telemetry API

orishoshan:
### Description

This PR removes the dependency on the Bugsnag SDK for error telemetry and uses a GraphQL-based API to report errors, similar to the usage telemetry API.


src/shared/telemetries/errorreporter/errorsender.go
```diff
@@ -0,0 +1,66 @@
...
+
+func newGqlClient() graphql.Client {
+	apiAddress := viper.GetString(telemetriesconfig.TelemetryAPIAddressKey)
+	clientTimeout := viper.GetDuration(telemetriesconfig.TimeoutKey)
+	clientWithTimeout := &http.Client{Timeout: clientTimeout}
+	return graphql.NewClient(apiAddress, clientWithTimeout)
+}
+
+func (s *ErrorSender) SendSync(errs []*telemetriesgql.Error) error {
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
NetanelBollag, line 31:
```suggestion
	clientTimeout := viper.GetDuration(telemetriesconfig.TimeoutKey)
	ctxTimeout, cancel := context.WithTimeout(context.Background(), clientTimeout)
```

src/shared/telemetries/errorreporter/errorsender.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func newGqlClient() graphql.Client {
+	apiAddress := viper.GetString(telemetriesconfig.TelemetryAPIAddressKey)
+	clientTimeout := viper.GetDuration(telemetriesconfig.TimeoutKey)
+	clientWithTimeout := &http.Client{Timeout: clientTimeout}
+	return graphql.NewClient(apiAddress, clientWithTimeout)
+}
+
+func (s *ErrorSender) SendSync(errs []*telemetriesgql.Error) error {
```
NetanelBollag, line 30: Guess we should also check here 
```suggestion
func (s *ErrorSender) SendSync(errs []*telemetriesgql.Error) error {
	if !s.enabled {
		return nil
	}

```

src/shared/telemetries/telemetriesconfig/config.go
```diff
@@ -7,6 +7,7 @@ import (
 
 const (
 	TelemetryAPIAddressKey         = "telemetry-address"
+	TelemetryAPIAddressDefault     = "http://localhost:8081/api/telemetry/query"
```
NetanelBollag, line 10: Doesn't this make deployments without helm charts disabled by default? Is there a way to config that only if run from the IDE or something

orishoshan: Fixed the config, thanks. The address no longer changes client-side, instead the server handles determining the release stage based on the version.

NetanelBollag (approved):
🚀

PR #437, opened 2024-05-28
Fix bug where DNS egress rules would be added twice; enable pprof by default on loopback

orishoshan:
### Description

Fixed a bug where the DNS egress rules created by the operator would be added twice to each network policy.

Also enabled pprof for the operator by default, to enable performance profiling. 
### Testing


src/operator/controllers/intents_reconcilers/networkpolicy/builders/dns_egress_network_policy.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func NewDNSEgressNetworkPolicyBuilder() *DNSEgressNetworkPolicyBuilder {
+	return &DNSEgressNetworkPolicyBuilder{}
+}
+
+// a function that creates []NetworkPolicyEgressRule from ep
+func (r *DNSEgressNetworkPolicyBuilder) buildNetworkPolicyEgressRules(ep effectivepolicy.ServiceEffectivePolicy) []v1.NetworkPolicyEgressRule {
+	if lo.NoneBy(ep.Calls, func(call otterizev1alpha3.Intent) bool {
+		return call.Type == "" || call.Type == otterizev1alpha3.IntentTypeHTTP || call.Type == otterizev1alpha3.IntentTypeKafka
```
omris94, line 26: you can use call.IsTargetInCluster()
```suggestion
		return call.IsTargetInCluster()
```

omris94 (approved):

PR #439, opened 2024-06-09
Promote the CRD API version `v1alpha3` to `v1`; introduce new CRD API version `v2alpha1` with a new structure of the `clientIntents`

omris94:
### Description

In this PR we:

- promote `v1alpha3` CRD version to `v1` (`v1alpha3` is still usable). 
- Create new API version `v2alpha1` with a new clientIntents format

### References

Include any links supporting this change such as a:

- GitHub Issue/PR number addressed or fixed
- StackOverflow post
- Related pull requests/issues from other repos

If there are no references, simply delete this section.

### Testing

### Checklist


src/operator/webhooks/clientintents_webhook_v2alpha1.go
```diff
@@ -0,0 +1,448 @@
...
+
+	if err := v.validateInternetTarget(intentTarget.Internet); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Validate that one and only one of the target's fields is set. Wwe do it by marshal the target to json and check there's only one key
+func (v *IntentsValidatorV2alpha1) validateOnlyOneTargetFieldSet(target otterizev2alpha1.Target) *field.Error {
```
omris94, line 212: @orishoshan - I feel like there must be a smarter way to do it, please tell me if something comes to your mind. 
(The only requirement: if a new "target type" is introduced (rabbitMQ for example), this test will function without any changes)

src/operator/config/webhook/manifests.yaml
```diff
@@ -32,6 +33,7 @@ webhooks:
       namespace: system
       path: /validate-k8s-otterize-com-v1alpha3-clientintents
   failurePolicy: Fail
+  matchPolicy: Exact
```
omris94, line 36: @orishoshan - Without `match:Policy: Exact` the client ignores the `apiVersion` rule - so in our case, each ClientIntent is sent to every validation webhook

src/operator/webhooks/postgresqlserverconfigs_webhook_v2alpha1.go
```diff
@@ -0,0 +1,106 @@
...
+
+var _ webhook.CustomValidator = &PostgresConfValidatorV2alpha1{}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (v *PostgresConfValidatorV2alpha1) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	return nil, nil
+}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (v *PostgresConfValidatorV2alpha1) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
```
amitlicht, line 61: make sure you port the changes made in this area in main (for PGServerConf & for MySQLServerConf)

amitlicht (changes requested):

src/operator/api/v2alpha1/clientintents_types.go
```diff
@@ -0,0 +1,888 @@
...
+}
+
+func (in *ClientIntents) GetCallsList() []Target {
+	return in.Spec.Targets
+}
+
+func (in *ClientIntents) GetFilteredCallsList(intentTypes ...IntentType) []Target {
+	return lo.Filter(in.GetCallsList(), func(item Target, index int) bool {
+		for _, intentType := range intentTypes {
+			if intentType == IntentTypeHTTP {
```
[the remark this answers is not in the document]
omris94: It doesn't translate to the same logic, because we iterate over the intent types and want to return only if there's a match (return true or continue). In this case, I have to use an extra `if` inside every case (for the nil check). I think that it is clearer when using `if` statements 🤔

src/operator/api/v1alpha3/webhooks.go
```diff
@@ -47,4 +146,152 @@ func (in *ClientIntents) SetupWebhookWithManager(mgr ctrl.Manager, validator web
...
+	dst := dstRaw.(*v2alpha1.ClientIntents)
+	dst.ObjectMeta = in.ObjectMeta
+	if dst.Spec == nil {
+		dst.Spec = &v2alpha1.IntentsSpec{}
+	}
+	dst.Spec.Workload.Name = in.Spec.Service.Name
+	dst.Spec.Workload.Kind = in.Spec.Service.Kind
+	dst.Spec.Targets = make([]v2alpha1.Target, len(in.Spec.Calls))
+	for i, call := range in.Spec.Calls {
+		if call.IsTargetInCluster() && call.Type != IntentTypeKafka {
```
amitlicht, line 159: I went and looked what `IntentTypeHTTP` means, and saw that it means Type is in {"", IntentTypeHTTP, IntentTypeKafka}. 
And then this code exclude IntentTypeKafka. 
So really, in this case it would just be clearer to replace this line with: 
`if call.Type == "" || call.Type == IntentTypeHTTP`. 

And then, since 2 lines from how you split again to the case of IntentTypeHTTP, how about just:
```
if call.Type == "" {
    kubernetesTarget := v2alpha1.KubernetesTarget{Name: call.GetServerFullyQualifiedName(in.Namespace), Kind: call.GetTargetServerKind()}
    dst.Spec.Targets[i] = v2alpha1.Target{Kubernetes: &kubernetesTarget}
}
if call.Type == HTTP { 
  ....
}
```

src/operator/api/v2alpha1/clientintents_types.go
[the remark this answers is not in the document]
amitlicht: it's not exactly the case, you're not looking for "return true or continue" because each case has a unique intent type. So switching this to switch by type and for each type return `item.X != nil` will have the same meaning. 

That said, leaving this discussion as I think it's a matter of taste and not crucial in any way. Fell free to resolve this comment and move on.

amitlicht (approved):

PR #442, opened 2024-06-23
Fixing bug where hashed usernames would exceed allowed MySQL username length

evyatarmeged:
### Description
Fixing a bug where hashed usernames would exceed allowed MySQL username length.

### Testing
Added a UT in database reconciler tests to ensure long workload & namespace names do not exceed the maximum length.

### Checklist

### DIFF
--- a/src/shared/clusterutils/username.go
+++ b/src/shared/clusterutils/username.go
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	HashedUsernameSectionMaxLength = 11
+	HashedUsernameSectionMaxLength = 10
 )
 
 func BuildHashedUsername(workloadName, namespace, clusterUID string) string {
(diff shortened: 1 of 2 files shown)


src/shared/clusterutils/username.go
```diff
@@ -31,9 +31,8 @@ func BuildHashedUsername(workloadName, namespace, clusterUID string) string {
 
 // KubernetesToPostgresName translates a name with Kubernetes conventions to Postgres conventions
 func KubernetesToPostgresName(kubernetesName string) string {
-	// '.' are replaced with dunders '__'
-	// '-' are replaced with single underscores '_'
-	return strings.ReplaceAll(strings.ReplaceAll(kubernetesName, ".", "__"), "-", "_")
+	// '.' and '-' are replaced with single underscores '_'
+	return strings.ReplaceAll(strings.ReplaceAll(kubernetesName, ".", "_"), "-", "_")
```
evyatarmeged, line 35: Note to reviewer - I changed this since it the replacing of `.` in `__` could also cause the name to exceed the allowed length.
This was originally here to allow to convert back from Postgres to K8s usernames, but we never used it.
I guess we'll have to deal it if we need it in the future (other ideas are welcome).

amitlicht (approved):

PR #446, opened 2024-06-27
Support enabling access to servers using pod annotation

NetanelBollag:
### Description

Support enabling access to server using pod annotation, for edge cases where applying client intents aren't possible 
In case the annotation intents are removed and there are no other client intents to the server, the operator might not handle the deletion event on time and the corresponding network policy won't get deleted. In that case, a quick solution might be restarting the operator pod or manually deleting the network policy. 

### Testing


src/operator/controllers/access_annotation/intents.go
```diff
@@ -0,0 +1,104 @@
...
+	otterizev1alpha3 "github.com/otterize/intents-operator/src/operator/api/v1alpha3"
+	"github.com/otterize/intents-operator/src/shared/errors"
+	"github.com/otterize/intents-operator/src/shared/injectablerecorder"
+	"github.com/otterize/intents-operator/src/shared/serviceidresolver/serviceidentity"
+	"github.com/samber/lo"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Intent struct {
```
omris94, line 14: Please consider changing this name because we already have several "intent" structs :-)

src/operator/controllers/access_annotation/parser.go
```diff
@@ -0,0 +1,81 @@
...
+
+func parseName(name string, kind string) (string, string, string, error) {
+	parts := strings.Split(name, ".")
+	if len(parts) != 2 {
+		return "", "", "", errors.Wrap(fmt.Errorf("bad called-by annotation client name '%s' should be in the format 'name.namespace'", name))
+	}
+	name = parts[0]
+	namespace := parts[1]
+
+	if len(kind) == 0 {
```
omris94, line 74: We do support kind OtterizLegacy so it is possible to add it here if you want

src/operator/controllers/access_annotation/intents.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		return AnnotationIntents{}, errors.Wrap(err)
+	}
+
+	for _, pod := range podList.Items {
+		if pod.GetDeletionTimestamp() != nil {
+			continue
+		}
+
+		clients, ok, err := ParseAccessAnnotations(&pod)
```
omris94, line 52: Maybe we should send an event to the pod if the parse failed?

src/operator/controllers/access_annotation/intents.go
NetanelBollag: Done

src/operator/controllers/access_annotation/parser.go
NetanelBollag: If Kind field is specified to Legacy everything should work just like in any other Kind

src/operator/controllers/access_annotation/intents.go
NetanelBollag: The error is recorded by the pods reconciler already, so no need to double recored it here, but I add a comment to make it clear

omris94 (approved):

PR #452, opened 2024-07-07
Add support for external traffic network policies that are restricted to only allow the Ingress controller, rather than all traffic

orishoshan:
### Description

Prior to this PR, all traffic would be allowed to a service whenever there was an Ingress or Service of type LoadBalancer or NodePort pointing to it (only on the relevant port). This meant that you could not block unauthorized access within the cluster. The network policies that allowed all traffic are called "external traffic network policies".

This PR makes it possible to restrict these external traffic network policies so that they only allow access from the ingress controller, in the event that the ingress controller's load balancer is an in-cluster pod. In some cases this is not relevant, as the load balancer could be a managed service outside of the cluster, the IP for which is not known so cannot be allowed selectively (e.g. such as when using AWS ELB).

This only applies when there is an Ingress and the traffic points to a Service with type ClusterIP. If the Service is of type LoadBalancer or NodePort, then all traffic is still allowed as before, in order to allow access from outside the cluster.

If no configuration is specified, all traffic is still allowed as before.

Example of a network policy that is restricted (note the `spec.ingress.from` section):
```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  annotations:
    intents.otterize.com/created-for-ingress: demo-ingress
    intents.otterize.com/created-for-service: server
  labels:
    intents.otterize.com/network-policy-external-traffic: server-otterize-tutorial-ia-68873b
  name: external-access-to-server
  namespace: otterize-tutorial-iam
  ownerReferences:
  - apiVersion: v1
    kind: Service
    name: server
spec:
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          kubernetes.io/metadata.name: ingress-nginx
      podSelector:
        matchLabels:
          intents.otterize.com/owner-kind: Deployment
          intents.otterize.com/service: ingress-nginx-contro-ingress-nginx-21081a
    ports:
    - port: 80
      protocol: TCP
  podSelector:
    matchLabels:
      intents.otterize.com/service: server-otterize-tutorial-ia-68873b
  policyTypes:
  - Ingress
```

This is a result of the following configuration, in the config.yaml (which is normally configured using Helm):
```
ingressControllers:
 name: ingress-nginx-controller
 namespace: ingress-nginx
 kind: Deployment
```

### Testing

### Checklist


src/shared/otterizecloud/graphqlclient/generate.go
```diff
@@ -2,5 +2,5 @@ package graphqlclient
 
 import _ "github.com/suessflorian/gqlfetch"
 
-//go:generate sh -c "go run github.com/suessflorian/gqlfetch/gqlfetch --endpoint https://app.staging.otterize.com/api/graphql/v1beta > schema.graphql"
```
amitlicht, line 5: revert this back to staging so that building the operator will work for users who aren't running the cloud locally.

amitlicht (approved):

PR #476, opened 2024-08-25
Add support for detecting AWS ALB ingress resources and auto-allowing traffic

orishoshan:
Enable with `--set operator.ingressControllerAWSALBExempt=true`. Use `intentsOperator.operator.ingressControllerAWSALBExempt` if using the `otterize-kubernetes` Helm chart.

### Description

Prior to this PR, if the `ingressControllers` configuration was passed in order to restrict Ingress traffic to in-cluster ingress Pods, AWS ALB would be (correctly) blocked. This PR enables the intents operator to detect when an AWS ALB ingress resource is used in tandem, and to open access to that pod as required.

### DIFF
--- a/src/operator/controllers/intents_reconcilers/external_traffic_network_policy/external_traffic_network_policy_test.go
+++ b/src/operator/controllers/intents_reconcilers/external_traffic_network_policy/external_traffic_network_policy_test.go
@@ -92,7 +92,7 @@ func (s *ExternalNetworkPolicyReconcilerTestSuite) SetupTest() {
 	testName := s.T().Name()
 	isShadowMode := strings.Contains(testName, "ShadowMode")
 	defaultActive := !isShadowMode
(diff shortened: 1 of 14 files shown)


src/shared/otterizecloud/graphqlclient/generate.go
```diff
@@ -2,5 +2,5 @@ package graphqlclient
 
 import _ "github.com/suessflorian/gqlfetch"
 
-//go:generate sh -c "go run github.com/suessflorian/gqlfetch/gqlfetch --endpoint https://app.staging.otterize.com/api/graphql/v1beta > schema.graphql"
```
omris94, line 5: why?

src/operator/controllers/intents_reconcilers/external_traffic_network_policy/external_traffic_network_policy_test.go
omris94, line 1: Add test please 🙏

omris94 (approved):

PR #480, opened 2024-08-29
Use `"alb.ingress.kubernetes.io/target-type"` instead of `"alb.ingress.kubernetes.io/scheme"` to detect AWS ALB presence on Ingress

orishoshan:
This was done because `target-type`: `ip` is sufficient to indicate that an AWS ALB will attempt to connect, rather than `scheme`: `internet-facing`.

### DIFF
--- a/src/operator/controllers/external_traffic/shared.go
+++ b/src/operator/controllers/external_traffic/shared.go
@@ -25,17 +25,17 @@ func serviceNamesFromIngress(ingress *v1.Ingress) sets.Set[string] {
 	return serviceNames
 }
 
-func isIngressListHasInternetFacingAWSALB(ingressList []v1.Ingress) bool {
+func isIngressListHasIPAWSALB(ingressList []v1.Ingress) bool {
 	return lo.SomeBy(ingressList, func(ingress v1.Ingress) bool {
 		if ingress.Annotations == nil {
 			return false
 		}
 
-		scheme, ok := ingress.Annotations["alb.ingress.kubernetes.io/scheme"]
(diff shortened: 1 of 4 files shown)


src/operator/controllers/external_traffic/shared.go
```diff
@@ -25,17 +25,17 @@ func serviceNamesFromIngress(ingress *v1.Ingress) sets.Set[string] {
...
 
-func isIngressListHasInternetFacingAWSALB(ingressList []v1.Ingress) bool {
+func isIngressListHasIPAWSALB(ingressList []v1.Ingress) bool {
 	return lo.SomeBy(ingressList, func(ingress v1.Ingress) bool {
 		if ingress.Annotations == nil {
 			return false
 		}
 
-		scheme, ok := ingress.Annotations["alb.ingress.kubernetes.io/scheme"]
+		scheme, ok := ingress.Annotations["alb.ingress.kubernetes.io/target-type"]
```
omris94, line 34:
```suggestion
		targetType, ok := ingress.Annotations["alb.ingress.kubernetes.io/target-type"]
```

src/operator/controllers/external_traffic/shared.go
```diff
@@ -25,17 +25,17 @@ func serviceNamesFromIngress(ingress *v1.Ingress) sets.Set[string] {
...
 		if !ok {
 			return false
 		}
 
-		return scheme == "internet-facing"
+		return scheme == "ip"
```
omris94, line 39:
```suggestion
		return targetType == "ip"
```

omris94 (approved):

PR #487, opened 2024-09-18
Fix bug where "external allow" policies were created if "AllowExternalTraffic" was set to "Always" even though the enforcement was disabled

omris94:
### Description

Fix bug where "external allow" policies were created when the enforcement was disabled and "AllowExternalTraffic" was set to "Always"

### Testing

### Checklist

### DIFF
--- a/src/operator/controllers/external_traffic/network_policy.go
+++ b/src/operator/controllers/external_traffic/network_policy.go
@@ -7,6 +7,7 @@ import (
 	"github.com/otterize/intents-operator/src/shared/errors"
 	"github.com/otterize/intents-operator/src/shared/injectablerecorder"
 	"github.com/otterize/intents-operator/src/shared/operatorconfig/allowexternaltraffic"
+	"github.com/otterize/intents-operator/src/shared/operatorconfig/enforcement"
 	"github.com/otterize/intents-operator/src/shared/serviceidresolver"
 	"github.com/otterize/intents-operator/src/shared/serviceidresolver/serviceidentity"
 	"github.com/samber/lo"
@@ -45,6 +46,7 @@ type NetworkPolicyHandler struct {
 	allowExternalTraffic         allowexternaltraffic.Enum
 	ingressControllerIdentities  []serviceidentity.ServiceIdentity
 	ingressControllerALBAllowAll bool
+	enforcementConfig            enforcement.Config
 }
 
 func NewNetworkPolicyHandler(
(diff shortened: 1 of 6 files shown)


src/shared/operatorconfig/enforcement/config.go
```diff
@@ -17,6 +18,20 @@ type Config struct {
...
+	if c.AllowExternalTraffic == allowexternaltraffic.Off {
+		return allowexternaltraffic.Off
+	} else if !c.EnforcementDefaultState && c.AllowExternalTraffic == allowexternaltraffic.Always {
+		// We don't want to create network policies for external traffic when enforcement is disabled.
+		// However, if one uses shadow mode we can still block external traffic to his protected services
+		// therefore we should return allowexternaltraffic.IfBlockedByOtterize
+		return allowexternaltraffic.IfBlockedByOtterize
+	} else {
+		return c.AllowExternalTraffic
+	}
```
amitlicht, line 34: consider using switch, plus this would be less confusing IMO:
```
switch c.AllowExternalTraffic {
case allowexternaltraffic.Off: return allowexternaltraffic.Off
case allowexternaltraffic.Always:
    if c.EnforcementDefaultState { return allowexternaltraffic.Always }
    else {return allowexternaltraffic.IfBlockedByOtterize} // and the long documentation goes here
default: // allowexternaltraffic.IfBlockedByOtterize
   return allowexternaltraffic.IfBlockedByOtterize
}
}
```

src/shared/operatorconfig/enforcement/config.go
```diff
@@ -75,5 +95,10 @@ func GetConfig() Config {
 		EnableGCPPolicy:                      viper.GetBool(EnableGCPPolicyKey),
 		EnableAzurePolicy:                    viper.GetBool(EnableAzurePolicyKey),
 		EnforcedNamespaces:                   goset.FromSlice(viper.GetStringSlice(ActiveEnforcementNamespacesKey)),
+		AllowExternalTraffic:                 allowexternaltraffic.Enum(viper.GetString(AllowExternalTrafficKey)),
 	}
 }
+
+const AllowExternalTrafficKey = "allow-external-traffic" // Whether to automatically create network policies for external traffic
+
+const AllowExternalTrafficDefault = allowexternaltraffic.IfBlockedByOtterize
```
amitlicht, line 104: move this to where the other consts are, at the top of the file

src/operator/controllers/external_traffic/network_policy.go
```diff
@@ -54,7 +56,14 @@ func NewNetworkPolicyHandler(
...
 	ingressControllerALBAllowAll bool,
 ) *NetworkPolicyHandler {
-	return &NetworkPolicyHandler{client: client, scheme: scheme, allowExternalTraffic: allowExternalTraffic, ingressControllerIdentities: ingressControllerIdentities, ingressControllerALBAllowAll: ingressControllerALBAllowAll}
+	return &NetworkPolicyHandler{
+		client:                       client,
+		scheme:                       scheme,
+		allowExternalTraffic:         allowExternalTraffic,
+		ingressControllerIdentities:  ingressControllerIdentities,
+		ingressControllerALBAllowAll: ingressControllerALBAllowAll,
+		enforcementConfig:            enforcement.GetConfig(),
```
amitlicht, line 65: since you're already querying the enforcementConfig here, no point in passing in the allowExternalTraffic value as argument (it's part of the enforcement config). Remove the input argument and just use enforcementConfig all around.

amitlicht (commented):
see comments

amitlicht (approved):

src/operator/controllers/external_traffic/network_policy.go
omris94: Actually, I don't use `enforcementConfig` I meant to do it but eventually kept the same signature. I will remove this line

amitlicht: IMO better use the enfocementConfig here than to input the specific flag(s), but it's a style thing, so you choose.

PR #488, opened 2024-09-21
Workaround AWS VPC CNI egress network policy agent bug: only allows traffic to ClusterIP if exact service selector is specified

orishoshan:
### Description

This works around an issue with the AWS VPC CNI where it does not allow traffic to the ClusterIP of a Service unless the NetworkPolicy uses the exact selector used by the Service.

Also fixes an unrelated issue where DNS traffic was not automatically allowed if the only targets were Internet intents.

This Pod has two sets of labels:
```
❯ kubectl get pods -n otterize-tutorial-iam -l intents.otterize.com/service=server-otterize-tutorial-ia-68873b
NAME                      READY   STATUS    RESTARTS   AGE
server-6dbbc4666f-nncjf   1/1     Running   0          85m
 ❯ kubectl get pods -n otterize-tutorial-iam -l app=server
NAME                      READY   STATUS    RESTARTS   AGE
server-6dbbc4666f-nncjf   1/1     Running   0          85m
```
app=server is the one used by the Service.
This network policy doesn't work:
```
  egress:
  - to:
    - namespaceSelector:
        matchLabels:
          kubernetes.io/metadata.name: otterize-tutorial-iam
      podSelector:
        matchLabels:
          intents.otterize.com/service: server-otterize-tutorial-ia-68873b
```
The traffic is denied. The VPC CNI log shows that the traffic is to the ClusterIP of the Service, rather than directly to the pod (which makes sense, but it should not be denied, since the correct Pod is selected).
```
{"level":"info","ts":"2024-09-20T12:55:01.282Z","logger":"ebpf-client","caller":"events/events.go:193","msg":"Flow Info:  ","Src IP":"192.168.169.89","Src Port":46652,"Dest IP":"10.100.204.247","Dest Port":80,"Proto":"TCP","Verdict":"DENY"}
```
If we also add this rule to the network policy, it starts working, because it is the exact label used by the Service with the ClusterIP 10.100.204.247. It looks like they don't correctly resolve Services, unless the exact label used by the Service is used.

```
  egress:
  - to:
    - namespaceSelector:
        matchLabels:
          kubernetes.io/metadata.name: otterize-tutorial-iam
      podSelector:
        matchLabels:
          app: server
```

```
"level":"info","ts":"2024-09-20T13:16:54.050Z","logger":"ebpf-client","caller":"events/events.go:193","msg":"Flow Info:  ","Src IP":"192.168.169.89","Src Port":37750,"Dest IP":"10.100.204.247","Dest Port":80,"Proto":"TCP","Verdict":"ACCEPT"}
```

### References

AWS support case 172678650700597. Will be updated with link if made public by AWS.

### Testing

### DIFF
--- a/src/shared/operator_cloud_client/intent_events_sender.go
+++ b/src/shared/operator_cloud_client/intent_events_sender.go
@@ -82,7 +82,7 @@ func (ies *IntentEventsPeriodicReporter) Start(ctx context.Context) error {
 	go func() {
 		defer errorreporter.AutoNotify()
 		// Wait for caches to sync
-		ies.waitForCacheSync(ctx)
+		ies.k8sClusterManager.GetCache().WaitForCacheSync(ctx)
 
 		ies.startReportLoop(ctx)
 	}()
@@ -106,21 +106,6 @@ func (ies *IntentEventsPeriodicReporter) startReportLoop(ctx context.Context) {
 		case <-statusReportTicker.C:
 			ies.reportIntentStatuses(ctx)
 		}
-	}
-}
-
-func (ies *IntentEventsPeriodicReporter) waitForCacheSync(ctx context.Context) {
-	for {
-		ok := func() bool {
-			timeoutCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
-			defer cancel()
-			return ies.k8sClusterManager.GetCache().WaitForCacheSync(timeoutCtx)
-		}()
-		if !ok {
-			logrus.Error("Intents Event Sender - Failed waiting for caches to sync")
-			continue
-		}
-		break
 	}
 }
 
--- a/src/shared/serviceidresolver/serviceidentity/serviceidentity.go
+++ b/src/shared/serviceidresolver/serviceidentity/serviceidentity.go
@@ -25,34 +25,34 @@ type ServiceIdentity struct {
 const KindService = "Service"
 const KindOtterizeLegacy = "OttrLegacy"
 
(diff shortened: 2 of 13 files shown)


src/shared/operator_cloud_client/intent_events_sender.go
```diff
@@ -82,7 +82,7 @@ func (ies *IntentEventsPeriodicReporter) Start(ctx context.Context) error {
 	go func() {
 		defer errorreporter.AutoNotify()
 		// Wait for caches to sync
-		ies.waitForCacheSync(ctx)
+		ies.k8sClusterManager.GetCache().WaitForCacheSync(ctx)
```
orishoshan, line 85: Changed this to try once, as this waits until cache sync, and can't be retried (once it succeeds it returns immediately, once it fails it fails immediately - causes log spam)

src/shared/serviceidresolver/serviceidentity/serviceidentity.go
```diff
@@ -25,34 +25,34 @@ type ServiceIdentity struct {
 const KindService = "Service"
 const KindOtterizeLegacy = "OttrLegacy"
 
-func (si *ServiceIdentity) GetFormattedOtterizeIdentityWithoutKind() string {
+func (si ServiceIdentity) GetFormattedOtterizeIdentityWithoutKind() string {
```
orishoshan, line 28: Fixes an issue where this couldn't be called in some context because the object wasn't a pointer

omris94 (approved):

PR #489, opened 2024-09-22
Bugfix in AWS IAM Roles Anywhere: last page of IAM roles anywhere profiles was not loaded into cache

amitlicht:
### Description
This PR fixes a bug in AWS IAM Roles Anywhere integration: when loading the Roles Anywhere profiles, the last page of profiles was not stored into the in-memory cache. As a result, the credentials-operator would attempt to re-create profiles for roles that already have profiles, causing an error: "provided role ARN should be unique across all profiles". 

### Testing

### Checklist

### DIFF
--- a/src/shared/awsagent/roles.go
+++ b/src/shared/awsagent/roles.go
@@ -51,6 +51,18 @@ func (a *Agent) GetOtterizeRole(ctx context.Context, namespaceName, accountName 
 	return true, role.Role, nil
 }
 
+func (a *Agent) loadNextProfileCachePage(ctx context.Context, nextToken *string) (*string, error) {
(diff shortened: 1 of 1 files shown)


src/shared/awsagent/roles.go
```diff
@@ -59,25 +71,15 @@ func (a *Agent) initProfileCache(ctx context.Context) (err error) {
...
 
-		for _, profile := range output.Profiles {
-			a.profileNameToId[*profile.Name] = *profile.ProfileId
-		}
-
-		for output.NextToken != nil {
-			nextToken = output.NextToken
-			for _, profile := range output.Profiles {
-				a.profileNameToId[*profile.Name] = *profile.ProfileId
-			}
```
amitlicht, line 77: Note for reviewer: the bug was that this line was not called on the last chunk of the iteration (when `nextToken = nil` again).

omris94 (approved):

PR #498, opened 2024-10-05
Handling some edge cases for network policy creation errors that are legitamate and should not break reconciliation  

evyatarmeged:
### Description
Handling some rare (but still possible) cases we have seen lately:
* `AlreadyExists` error on netpol creation
* `Forbidden` for namespace being terminated
* Missing namespace (was deleted before .`Create()` was called)


src/operator/controllers/intents_reconcilers/networkpolicy/reconciler.go
```diff
@@ -474,6 +475,45 @@ func (r *Reconciler) reconcileEndpointsForPolicy(ctx context.Context, newPolicy
...
+	err := r.Get(ctx, types.NamespacedName{
+		Name:      netpol.Name,
+		Namespace: ep.Service.Namespace},
+		existingPolicy)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		r.RecordWarningEventf(existingPolicy, consts.ReasonGettingNetworkPolicyFailed, "failed to get network policy: %s", err.Error())
+		return errors.Wrap(err)
+	}
+	if k8serrors.IsNotFound(err) {
+		return errors.Wrap(fmt.Errorf("failed creating network policy with AlreadyExists err, but failed getting netpol from k8s api server"))
```
amitlicht, line 489: use `errors.Errorf`

src/operator/controllers/intents_reconcilers/networkpolicy/reconciler.go
```diff
@@ -474,6 +475,45 @@ func (r *Reconciler) reconcileEndpointsForPolicy(ctx context.Context, newPolicy
...
+		return r.handleExistingPolicyRetry(ctx, ep, netpol)
+	}
+
+	if k8serrors.IsForbidden(err) && strings.Contains(errStr, "is being terminated") {
+		// Namespace is being deleted, nothing to do further
+		logrus.Debugf("Namespace %s is being terminated, ignoring api server error", netpol.Namespace)
+		return nil
+	}
+
+	if strings.Contains(errStr, netpol.Namespace) && strings.Contains(errStr, "not found") {
```
amitlicht, line 508: should we check the error type here? (should be NotFound error)

amitlicht (approved):
see comments

src/operator/controllers/intents_reconcilers/networkpolicy/reconciler.go
evyatarmeged: Wasn't sure if for missing namespace on Create that returning error type is `NotFound`, but it is. Changed :+1:

PR #501, opened 2024-10-13
Catch "already exists" error to avoid unnecessary error events

omris94:
### Description

Catch "already exists" error to avoid unnecessary error events

### DIFF
--- a/src/operator/controllers/istiopolicy/policy_manager.go
+++ b/src/operator/controllers/istiopolicy/policy_manager.go
@@ -402,10 +402,9 @@ func (c *PolicyManagerImpl) createOrUpdatePolicies(
 
 		err = c.client.Create(ctx, newPolicy)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


src/operator/controllers/istiopolicy/policy_manager.go
```diff
@@ -402,7 +402,7 @@ func (c *PolicyManagerImpl) createOrUpdatePolicies(
 
 		err = c.client.Create(ctx, newPolicy)
 		if err != nil {
-			if k8serrors.IsConflict(err) {
+			if k8serrors.IsConflict(err) || k8serrors.IsAlreadyExists(err) {
```
evyatarmeged, line 405: Shouldn't there be an `updateExisting` logic that we should fallback to ? With `Requeue: true` maybe ?
Asking based on similar logic in netpols creation, might be wrong :)

evyatarmeged (approved):

PR #517, opened 2024-11-25
Add support for azure custom roles

otterobert:
### Description

Add support for azure custom roles to enable fine-grained permission control instead of preset role assignments

### Testing

### Checklist


src/operator/controllers/intents_reconcilers/iam/iampolicyagents/azurepolicyagent/agent.go
```diff
@@ -301,3 +323,61 @@ func (a *Agent) vaultAccessPolicyEntryFromIntent(userAssignedIdentity armmsi.Ide
 		},
 	}
 }
+
+func (a *Agent) ensureCustomRolesForIntents(ctx context.Context, userAssignedIdentity armmsi.Identity, intents []otterizev2alpha1.Target) error {
+	// Lock the agent to ensure that no other goroutine is modifying the custom roles in parallel
+	a.roleMutex.Lock()
+	defer a.roleMutex.Unlock()
```
amitlicht, line 330: Is it enough to lock around ensureCustomRoleForIntent (single) instead? So that the lock is held for a shorter time.

src/shared/azureagent/customroles.go
```diff
@@ -0,0 +1,127 @@
...
+	resp, err := a.roleDefinitionsClient.CreateOrUpdate(ctx, roleScope, id, roleDefinition, nil)
+	if err != nil {
+		return errors.Wrap(err)
+	}
+
+	// create a role assignment for the custom role
+	err = a.CreateRoleAssignment(ctx, scope, uai, resp.RoleDefinition, to.Ptr(otterizeCustomRoleTag))
+	if err != nil {
+		return errors.Wrap(err)
+	}
```
amitlicht, line 70: The fact that this function is called `CreateCustomRole`, but in reality it creates and assigns, is confusing IMO. 
I'd split the `CreateRoleAssignment` out of it, and have the caller (agent code) call `CreateRoleAssignment` explicitly. 

Alternatively, rename the function to `CreateAndAssignCustomRole`

src/shared/azureagent/customroles.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return nil
+}
+
+func (a *Agent) UpdateCustomRole(ctx context.Context, role *armauthorization.RoleDefinition, actions []v2alpha1.AzureAction, dataActions []v2alpha1.AzureDataAction) error {
```
amitlicht, line 75: Potential bug: 
if`CreateCustomRole` was called and for some reason failed after `a.roleDefinitionsClient.CreateOrUpdate`, but before `a.CreateRoleAssignmen`, then -- on the next reconcile attempt, the custom role already exists but the role assignment does not. You would then call `UpdateCustomRole`, but the assignment will never be created. Is that true?

amitlicht (approved):

src/operator/controllers/intents_reconcilers/iam/iampolicyagents/azurepolicyagent/agent.go
otterobert: moved

PR #522, opened 2024-12-01
Handle azure iam hanging resources

otterobert:
### Description

Handle cases where there are hanging custom roles and assignments after deleting intents and kubernetes service accounts

### Testing

### Checklist

### DIFF
--- a/src/shared/azureagent/identities.go
+++ b/src/shared/azureagent/identities.go
@@ -99,8 +99,26 @@ func (a *Agent) DeleteUserAssignedIdentity(ctx context.Context, namespace string
 	userAssignedIdentityName := a.GenerateUserAssignedIdentityName(namespace, accountName)
 	federatedIdentityCredentialsName := a.generateFederatedIdentityCredentialsName(namespace, accountName)
 
+	// Delete roles assigned to the identity
+	identity, err := a.FindUserAssignedIdentity(ctx, namespace, accountName)
+	if err != nil {
+		return errors.Wrap(err)
+	}
+
+	roleAssignments, err := a.ListRoleAssignments(ctx, identity)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


src/shared/azureagent/identities.go
```diff
@@ -99,8 +99,26 @@ func (a *Agent) DeleteUserAssignedIdentity(ctx context.Context, namespace string
 	userAssignedIdentityName := a.GenerateUserAssignedIdentityName(namespace, accountName)
 	federatedIdentityCredentialsName := a.generateFederatedIdentityCredentialsName(namespace, accountName)
 
+	// Delete roles assigned to the identity
+	identity, err := a.FindUserAssignedIdentity(ctx, namespace, accountName)
+	if err != nil {
```
amitlicht, line 104: you should catch and ignore NotFound errors here, to avoid failing deletion just because the resource is already deleted. See reference a few lines down (`!azureerrors.IsNotFoundErr...`).

src/shared/azureagent/identities.go
```diff
@@ -99,8 +99,26 @@ func (a *Agent) DeleteUserAssignedIdentity(ctx context.Context, namespace string
...
+		return errors.Wrap(err)
+	}
+
+	roleAssignments, err := a.ListRoleAssignments(ctx, identity)
+	if err != nil {
+		return errors.Wrap(err)
+	}
+
+	for _, roleAssignment := range roleAssignments {
+		if err := a.DeleteRoleAssignment(ctx, roleAssignment); err != nil {
```
amitlicht, line 114: same here. In general, try to make this code as resilient as possible to missing resources. It shouldn't fail the reconciler just because some resource is already deleted.

amitlicht (approved):

PR #523, opened 2024-12-03
Validate scope before creating Azure custom roles

otterobert:
### Description

Validate scope before creating Azure custom roles - this will prevent cleanup errors and invalid role assignments

### Testing

### Checklist

### DIFF
--- a/src/shared/azureagent/customroles.go
+++ b/src/shared/azureagent/customroles.go
@@ -31,13 +31,14 @@ func (a *Agent) GenerateCustomRoleName(uai armmsi.Identity, scope string) string
 	return agentutils.TruncateHashName(fullName, maxRoleNameLength)
 }
 
-func (a *Agent) ValidateScope(ctx context.Context, scope string) bool {
+func (a *Agent) ValidateScope(ctx context.Context, scope string) error {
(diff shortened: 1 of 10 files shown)


src/shared/azureagent/customroles.go
```diff
@@ -31,6 +31,15 @@ func (a *Agent) GenerateCustomRoleName(uai armmsi.Identity, scope string) string
 	return agentutils.TruncateHashName(fullName, maxRoleNameLength)
 }
 
+func (a *Agent) ValidateScope(ctx context.Context, scope string) bool {
+	res, err := a.resourceClient.GetByID(ctx, scope, "2022-09-01", nil)
+	if err != nil {
+		return false
```
amitlicht, line 37: muting errors is generally bad practice, and can hide very bad things. 
Change it to a function returning error (instead of bool), so that `if err != nil { return errors.Wrap(err) };   else if res.GenericResource.ID != nil { return errors.Format(...not found) } ; else return nil`

That's also more in-line with the function name ("Validate" should validate something or fail, while function named "IsScopeExisting" would be expected to return a boolean)

amitlicht (approved):
approved pending comment inside

PR #529, opened 2024-12-10
Add support to partial azure identity scopes

otterobert:
### Description

Some Azure resources - such as storage accounts, are unique across azure and can be accessed directly without specifying subscription of resource group. This PR adds full scope resolution for partial scopes/ids to full scopes that include the subscription and resource group id.

### Testing

### Checklist


src/shared/azureagent/resources.go
```diff
@@ -0,0 +1,70 @@
...
+	"regexp"
+	"strings"
+)
+
+const StorageAccountPattern = `providers/Microsoft.Storage/storageAccounts/([^/]+)`
+
+func (a *Agent) ValidateScope(ctx context.Context, scope string) error {
+	res, err := a.resourceClient.GetByID(ctx, scope, "2022-09-01", nil)
+	if err != nil {
+		return err
```
amitlicht, line 16: errors.Wrap

src/shared/azureagent/resources.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Storage accounts are unique across azure, so we should look up the storage account directly
+	regex := regexp.MustCompile(StorageAccountPattern)
+	matchStorageAccount := regex.FindStringSubmatch(partialScope)
+	scopeToCheck := partialScope
+	if len(matchStorageAccount) > 1 {
+		scopeToCheck = fmt.Sprintf("/providers/Microsoft.Storage/storageAccounts/%s", matchStorageAccount[1])
+	}
+
+	for _, sub := range subscriptions {
```
amitlicht, line 38: This may take a really long time in a world where there's many subscriptions with many resources in them. 
Some ideas: 
1. use `resourceClient.NewListPager` with filter, to filter by resource type (storage accounts)
2. add some local cache and store resolved storage accounts to avoid searching for them over and over again

src/shared/azureagent/resources.go
```diff
@@ -0,0 +1,70 @@
...
+						// Rebuild the full scope:
+						// partialScope: <optional-prefix>/providers/Microsoft.Storage/storageAccounts/<storage-account-name>/<child-resource>
+						// scopeToCheck: /providers/Microsoft.Storage/storageAccounts/<storage-account-name>
+						// parts[1]: /<child-resource> || ""
+						parts := strings.Split(partialScope, scopeToCheck)
+						fullScope := fmt.Sprintf("%s%s", *resource.ID, parts[1])
+						return fullScope, nil
+					}
+					return *resource.ID, nil
+				}
```
amitlicht, line 64: move to a separate function to improve readability. Something like: 
```
resourceID, ok, err := a.findResourceInPage(...)
```

src/shared/azureagent/resources.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return err
+	}
+	if res.GenericResource.ID == nil {
+		return errors.Errorf("scope %s not found", scope)
+	}
+	return nil
+}
+
+func (a *Agent) GetFullResourceScope(ctx context.Context, partialScope string) (string, error) {
```
amitlicht, line 24: Actually, for any resource that's not a storage account, I wouldn't go into this lookup function. All other resources AFAIK do not define any cross-subscription uniqueness, so if, for example, our customers define a partial resource group scope `/resourceGroups/myResourceGroup`, there's nothing guaranteeing uniqueness cross subscription. 
Whats more, this can lead to breaking existing behavior. So far, on partial scopes, we assumed they'd belong to the same subscription and resource group as the AKS cluster. 

Instead, let's do this:
- if the partial scope is a storage account (or a container in it), lookup for it cross-subscriptions (by storage account name, which is globally unique). 
- otherwise, fallback to assuming its in the same subscription & resource group as the AKS cluster (keeping existing behavior).

amitlicht (changes requested):

amitlicht (approved):

PR #543, opened 2025-01-14
Report Istio installation status as Istio enforcement status

orishoshan:
This PR makes the intents operator report whether Istio is installed, which is automatically taken into account by Otterize Cloud to auto-hide the access statuses for Istio if not installed in the cluster.


evyatarmeged (approved):

orishoshan:
Will be merged as part of another branch

PR #551, opened 2025-01-22
Solve a potential race on AWS IAM integration where policy deletion would error because the related IAM role was not found

otterobert:
### Description

In cases when AWS roles were already deleted (possibly by the credentials operator) in a race condition, re-queue the event for another attempt.

### Checklist

### DIFF
--- a/src/operator/controllers/intents_reconcilers/iam/iam_reconciler.go
+++ b/src/operator/controllers/intents_reconcilers/iam/iam_reconciler.go
@@ -5,6 +5,7 @@ import (
 	otterizev2alpha1 "github.com/otterize/intents-operator/src/operator/api/v2alpha1"
 	"github.com/otterize/intents-operator/src/operator/controllers/intents_reconcilers/consts"
 	"github.com/otterize/intents-operator/src/operator/controllers/intents_reconcilers/iam/iampolicyagents"
+	"github.com/otterize/intents-operator/src/shared/awsagent"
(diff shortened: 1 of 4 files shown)


src/operator/controllers/intents_reconcilers/iam/iam_reconciler.go
```diff
@@ -83,6 +84,9 @@ func (r *IAMIntentsReconciler) Reconcile(ctx context.Context, req reconcile.Requ
 	}
 
 	if err := r.applyTypedIAMIntents(ctx, pod, intents, r.agent); err != nil {
+		if errors.Is(err, awsagent.ErrRoleNotFound) {
```
amitlicht, line 87: 1. We probably need to handle similar cases in AWS, Azure & GCP. In all of them, intent reconciliation may conflict with role/identity deletion. Please review the flows for the other IAM integrations as well. 
2. This part of the code is a "generic" IAM reconciler, so catching `awsagent.ErrRoleNotFound` error (that's specific to a certain IAM provider) seems unnatural. Maybe use a generalized `iampolicyagents.ErrRoleNotFound` that's going to be used for all 3 providers?

amitlicht, line 87: please add a UT case for this (under iam_reconciler_test.go)

amitlicht (changes requested):

amitlicht (approved):

PR #553, opened 2025-01-24
Periodically Check and cleanup orphaned resources in azure

otterobert:
### Description

Periodically Check and cleanup orphaned resources in azure that could have been left over because of network errors or race conditions. Some resources (such as custom roles and role assignments) must be deleted together and an interrupt in the middle of the deletion transaction might cause orphaned resources.

### Checklist

### DIFF
--- a/src/operator/controllers/intents_reconcilers/iam/iampolicyagents/azurepolicyagent/periodic.go
+++ b/src/operator/controllers/intents_reconcilers/iam/iampolicyagents/azurepolicyagent/periodic.go
@@ -1,16 +1,13 @@
-package azureagent
+package azurepolicyagent
 
 import (
 	"context"
-	"fmt"
 	"github.com/otterize/intents-operator/src/shared/errors"
 	"github.com/sirupsen/logrus"
 	"strings"
 )
 
-func (a *Agent) PeriodicTasks() {
-	ctx := context.TODO()
-
+func (a *Agent) PeriodicTasks(ctx context.Context) {
 	err := a.CleanupCustomRoles(ctx)
 	if err != nil {
 		logrus.WithError(err).Error("Failed to clean up custom roles")
@@ -20,10 +17,13 @@ func (a *Agent) PeriodicTasks() {
 // CleanupCustomRoles deletes custom roles that got orphaned due to the deletion of its role assignments
 // when deleting a role assignment, the custom role might not get deleted if some error occurs (DeleteRoleAssignment)
 func (a *Agent) CleanupCustomRoles(ctx context.Context) error {
+	// Lock the agent to ensure that no other goroutine is modifying the custom roles in parallel
+	a.roleMutex.Lock()
+	defer a.roleMutex.Unlock()
+
 	// List otterize custom roles across all subscriptions
 	existingCustomRoles, err := a.ListCustomRolesAcrossSubscriptions(ctx)
 	if err != nil {
-		fmt.Println("Error: ", err)
 		return errors.Wrap(err)
 	}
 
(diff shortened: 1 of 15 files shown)


src/shared/azureagent/periodic.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/otterize/intents-operator/src/shared/errors"
+	"github.com/sirupsen/logrus"
+	"strings"
+)
+
+func (a *Agent) PeriodicTasks() {
+	ctx := context.TODO()
```
amitlicht, line 12: Provide an external context for it, probably the one passed to `NewAzureAgent`. This is so that, if the context is canceled (eg by sigterm), the periodic tasks will gracefully terminate as well. 

You can do this: 
```
go wait.Forever(func() {agent.PeriodicTasks(ctx)}, 5*time.Hour)
```

src/shared/azureagent/roleassignments.go
```diff
@@ -93,6 +97,17 @@ func (a *Agent) ListRoleAssignmentsForSubscription(ctx context.Context, subscrip
 		}
 
 		for _, roleAssignment := range page.Value {
+			// We want to list only otterize role assignments
```
amitlicht, line 100: Then maybe rename the function to `ListOtterizeRoleAssignmentsForSubscription`?

src/shared/azureagent/periodic.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+// CleanupCustomRoles deletes custom roles that got orphaned due to the deletion of its role assignments
+// when deleting a role assignment, the custom role might not get deleted if some error occurs (DeleteRoleAssignment)
+func (a *Agent) CleanupCustomRoles(ctx context.Context) error {
+	// List otterize custom roles across all subscriptions
+	existingCustomRoles, err := a.ListCustomRolesAcrossSubscriptions(ctx)
+	if err != nil {
+		fmt.Println("Error: ", err)
```
amitlicht, line 26: remove the debug print

amitlicht: (or replace it with logrus.Error if you feel you need to)

src/shared/azureagent/periodic.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// List otterize role assignments across all subscriptions
+	existingRoleAssignments, err := a.ListRoleAssignmentsAcrossSubscriptions(ctx, nil)
+	if err != nil {
+		return errors.Wrap(err)
+	}
+
+	// Mark custom roles that are in use
+	for _, roleAssignment := range existingRoleAssignments {
+		if a.IsCustomRoleAssignment(roleAssignment) {
```
amitlicht, line 44: aren't you already doing this check in `ListRoleAssignmentsAcrossSubscriptions`?

src/shared/azureagent/periodic.go
```diff
@@ -0,0 +1,77 @@
...
+
+	err := a.CleanupCustomRoles(ctx)
+	if err != nil {
+		logrus.WithError(err).Error("Failed to clean up custom roles")
+	}
+}
+
+// CleanupCustomRoles deletes custom roles that got orphaned due to the deletion of its role assignments
+// when deleting a role assignment, the custom role might not get deleted if some error occurs (DeleteRoleAssignment)
+func (a *Agent) CleanupCustomRoles(ctx context.Context) error {
```
amitlicht, line 22: Can there be a race here? 
If there's a new intent being reconciled at the same time as the periodic check is happening, its role was _just_ created, and the role assignment still hasn't. Then the periodic task is going to try and delete the role, while the reconcile loop is trying to create the role assignment, and one of them will fail. 

One way to go about it is hold the roles mutex while the cleanup is happening...

src/shared/azureagent/roleassignments.go
otterobert: no. it's obvious we are working with otterize-specific roles and assignments - the comment is there to explain the filter without having the developer reference the function. if you don't like the comment i can remove it.

src/shared/azureagent/periodic.go
otterobert: si

src/shared/azureagent/periodic.go
otterobert: moved from periodic tasks form azureagent to azurepolicyagent and added the lock

src/operator/controllers/intents_reconcilers/iam/iampolicyagents/azurepolicyagent/periodic.go
```diff
@@ -0,0 +1,75 @@
...
+		rolesInUse[roleID] = true
+	}
+
+	// Delete custom roles that are not in use
+	for roleID, inUse := range rolesInUse {
+		if !inUse {
+			roleScope := a.GetSubscriptionScope(roleID)
+			roleDefinitionID := roleID[strings.LastIndex(roleID, "/")+1:]
+
+			err := a.DeleteCustomRole(ctx, roleScope, roleDefinitionID)
```
amitlicht, line 67: Add a `log.Warning` here about how you're deleting an orphan custom role. Just in case.

amitlicht (approved):

PR #561, opened 2025-02-12
Create warning event if client has internet intent without egress policy

zohar7ch:
### Description
If the egress policy is not enabled, we cannot create a network policy to protect this traffic, as any policy we create referring to the internet can only apply to the client, and thus, it would be an egress policy. 
Now, we will create warning to the ClientIntent to notify of this invalid configuration.

### Testing
Added a unit test that verifies that the warning event is being recorded.

### Checklist


.github/workflows/lint.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           working-directory: src
 
           # Optional: golangci-lint command line arguments.
-          args: --timeout 5m --out-format github-actions
+          args: --timeout 5m --out-format colored-line-number
```
zohar7ch, line 62: @omris94 otherwise I could not find the line where the error was.. though I not sure whether this is something we want to keep or not

PR #575, opened 2025-03-16
When creating an external traffic network policy, now only use a single existing Otterize network policy to determine the pod selector used by the external traffic network policy as all of the different pod selectors are equivalent

orishoshan:
### Description

This PR does not have any behavioral affect on the intents-operator. 
Before this change, when we had to create an external traffic network policy, we used to create one and modify it many times based on existing network policies selectors.
No we only choose one selector from one existing network policy, and create the network policy only once, thus reducing the number of operations we perform in the cluster.
It also makes the code more clear and easy to understand.

### Testing

### Checklist

### DIFF
--- a/src/operator/controllers/external_traffic/network_policy.go
+++ b/src/operator/controllers/external_traffic/network_policy.go
@@ -423,6 +423,8 @@ func (r *NetworkPolicyHandler) handleEndpointsWithIngressList(ctx context.Contex
 		netpolsAffectingThisWorkload = append(netpolsAffectingThisWorkload, netpolList.Items...)
 
 		// Sort netpolsAffectingThisWorkload by name so that we always use the same netpol when running Find below
+		// This is important because we relay on the order of the netpols to determine whether it already exists, does not
+		// exist or need an update
 		slices.SortStableFunc(netpolsAffectingThisWorkload, func(i, j v1.NetworkPolicy) bool {
 			return i.Name < j.Name
 		})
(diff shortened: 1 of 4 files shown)


src/operator/controllers/external_traffic/network_policy.go
```diff
@@ -395,35 +395,40 @@ func (r *NetworkPolicyHandler) handleEndpointsWithIngressList(ctx context.Contex
...
 		err = r.client.List(ctx, netpolList, client.MatchingLabels{v2alpha1.OtterizeNetworkPolicy: (&serviceidentity.ServiceIdentity{Name: endpoints.Name, Namespace: endpoints.Namespace, Kind: serviceidentity.KindService}).GetFormattedOtterizeIdentityWithKind()})
 		if err != nil {
 			return errors.Wrap(err)
 		}
 
-		netpolSlice = append(netpolSlice, netpolList.Items...)
+		netpolsAffectingThisWorkload = append(netpolsAffectingThisWorkload, netpolList.Items...)
 
-		hasIngressRules := lo.SomeBy(netpolSlice, func(netpol v1.NetworkPolicy) bool {
+		// Sort netpolsAffectingThisWorkload by name so that we always use the same netpol when running Find below
```
zohar7ch, line 426: maybe add here the reason for why this is important (something like: because later in the flow we compare the existing policy to the new one..)

zohar7ch (approved):

PR #586, opened 2025-04-19
Fixed bugs with creation & deletion of role assignments when using intents v1 Azure roles.

evyatarmeged:
### Description
Fixed bugs with creation & deletion of role assignments when using intents v1 Azure roles.
Slight refactoring and tweaks for how we handle roles & role assignments, especially deletion flows.
Removed `agent_periodic_test.go`, since periodic cleanups were removed.


src/shared/azureagent/roleassignments.go
```diff
@@ -134,13 +112,5 @@ func (a *Agent) FindRoleDefinitionByName(ctx context.Context, scope string, role
...
 	}
 
-	missingRoles := lo.Filter(roleNames, func(roleName string, _ int) bool {
-		_, exists := roleDefinitionsByName[roleName]
-		return !exists
-	})
-	if len(missingRoles) > 0 {
-		return nil, errors.Errorf("azure role definitions not found: %s", missingRoles)
-	}
-
```
evyatarmeged, line 144: @orishoshan Caused an error when tested, doesn't seem like it's needed with new logic (doesn't exist - move on)

src/shared/azureagent/agent.go
```diff
@@ -40,7 +40,7 @@ type Agent struct {
 	managedClustersClient               AzureARMContainerServiceManagedClustersClient
 	userAssignedIdentitiesClient        AzureARMMSIUserAssignedIdentitiesClient
 	federatedIdentityCredentialsClient  AzureARMMSIFederatedIdentityCredentialsClient
-	roleDefinitionsClient               AzureARMAuthorizationRoleDefinitionsClient
+	RoleDefinitionsClient               AzureARMAuthorizationRoleDefinitionsClient
```
orishoshan, line 43: Why is this public now? As far as I can tell it's not needed and can stay private?

src/operator/controllers/intents_reconcilers/iam/iampolicyagents/azurepolicyagent/agent_identities_test.go
orishoshan, line 1: It seems to me that these tests should not be deleted

orishoshan (commented):
See comments. And also, we need a full-flow tests, from ClientIntents to calling APIs, with both usages of the CRD (roles and actions). To be clear, it doesn't have to be a new test if the flow is already tested, but the test has to be failing on the old code and passing on the new code, and covering the scenarios that were fixed.

src/shared/azureagent/agent.go
evyatarmeged: IDE lost its mind. Reverted

src/operator/controllers/intents_reconcilers/iam/iampolicyagents/azurepolicyagent/agent.go
```diff
@@ -154,18 +144,29 @@ func (a *Agent) ensureRoleAssignmentsForIntents(ctx context.Context, userAssigne
...
 	existingRoleDefinitionIDs := goset.FromSlice(lo.Map(existingRoleAssignmentsForScope, func(roleAssignment armauthorization.RoleAssignment, _ int) string {
 		return *roleAssignment.Properties.RoleDefinitionID
 	}))
 
 	for _, roleName := range roleNames {
-		roleDefinition := roleDefinitionsByName[roleName]
+		roleDefinition, exists := roleDefinitionsByName[roleName]
+		if !exists {
+			// The role mentioned in the intents does not exist in the subscription, record event and move on
+			a.RecordWarningEventf(&intents, ReasonRoleDefinitionNotFound, "Role definition %s not found in scope %s", roleName, scope)
```
orishoshan, line 167:
```suggestion
			a.RecordWarningEventf(&intents, ReasonRoleNotFound, "Role %s not found in scope %s", roleName, scope)
```
From the user's perspective he's specifying roles, not role definitions (in the ClientIntents), so the error should use the same terminology

src/operator/controllers/intents_reconcilers/iam/iampolicyagents/azurepolicyagent/agent_customroles_test.go
```diff
@@ -316,21 +342,221 @@ func (s *AzureAgentPoliciesCustomRolesSuite) TestAddRolePolicyFromIntents_Custom
...
+
+	// Two calls - one from custom roles and one from backwards compatibility to built-in roles
+	s.expectListSubscriptionsReturnsPager()
+	s.expectListSubscriptionsReturnsPager()
+	s.expectListRoleAssignmentsReturnsEmpty()
+	s.expectListRoleAssignmentsReturnsEmpty()
+	s.expectListRoleDefinitionsReturnsPager(existingRoles) // Pager should return the roles from the intents
+	s.expectCreateRoleAssignmentReturnsEmpty()
+
+	// We expect s.mockRoleDefinitionsClient.EXPECT().CreateOrUpdate() to NOT be called - the roles already exist
```
orishoshan, line 396: Instead of this comment, add an EXPECT() with .Times(0)
Something like...
```suggestion
	s.mockRoleDefinitionsClient.EXPECT().CreateOrUpdate(gomock.Any(), ...).Times(0)
```

orishoshan (approved):

